PR #2, opened 2019-07-16
Feature/cp assessments

tsubik:
Adding carbon performance assessment

```
bundle exec rake import:all
```

image

### DIFF
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -113,7 +113,7 @@ ActiveAdmin.register Company do
                       tr do
                         a.emissions_all_years.map do |year|
                           td do
-                            a.emissions[year]
+                            a.emissions[year] || 'N/A'
                           end
                         end
                       end
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -86,28 +86,7 @@ ActiveAdmin.register Company do
                   row :assumptions
                 end
 
-                if a.emissions.present?
-                  table class: 'table cell-padding-sm cell-centered' do
-                    thead do
-                      a.emissions_all_years.map do |year|
-                        th year
-                      end
-                    end
-                    tbody do
-                      tr do
-                        a.emissions_all_years.map do |year|
-                          td do
-                            a.emissions[year] || 'N/A'
-                          end
-                        end
-                      end
-                    end
-                  end
-                else
-                  div class: 'padding-20' do
-                    'No emission or target values for this assessment'
-                  end
-                end
+                render 'admin/cp/emissions_table', emissions: a.emissions
               end
             end
           end
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -54,7 +54,9 @@ ActiveAdmin.register Company do
       tab :mq_assessments do
         panel 'Management Quality Assessments' do
           if resource.mq_assessments.empty?
-            'No Management Quality Assessments for this company yet'
+            div class: 'padding-20' do
+              'No Management Quality Assessments for this company yet'
+            end
           else
             resource.mq_assessments.latest_first.map do |a|
               panel "Assessement taken on #{a.assessment_date}", class: 'mq_assessment' do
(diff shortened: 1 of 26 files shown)


app/admin/companies.rb
```diff
@@ -54,14 +54,16 @@
...
+              'No Management Quality Assessments for this company yet'
+            end
           else
             resource.mq_assessments.latest_first.map do |a|
               panel "Assessement taken on #{a.assessment_date}", class: 'mq_assessment' do
                 attributes_table_for a do
                   row :level, &:status_description_short
                   row :publication_date do
-                    "#{a.publication_date.strftime('%B %Y')}"
+                    a.publication_date.strftime('%B %Y')
```
kowal: we could use rails date/time formats for this
https://api.rubyonrails.org/classes/Time.html#method-i-to_formatted_s

app/models/mq/assessment.rb
```diff
@@ -15,7 +15,7 @@
 
 module MQ
   class Assessment < ApplicationRecord
-    LEVELS = %w[1 2 3 4 4STAR].freeze
+    LEVELS = %w[0 1 2 3 4 4STAR].freeze
```
kowal: what does `4STAR` mean here?

app/admin/companies.rb
```diff
@@ -84,6 +86,48 @@
       end
 
       tab :cp_assessments do
+        panel 'Carbon Performance Assessments' do
+          if resource.cp_assessments.empty?
+            div class: 'padding-20' do
+              'No Carbon Performance Assessments for this company yet'
```
kowal: I think we could wrap resources in some kind of presenter object, which could handle things like:
- placeholders for whole resource/empty fields (`No .. yet`, `N/A`, ..)
- formatting dates

What do you think?

app/models/cp/assessment.rb
```diff
@@ -0,0 +1,15 @@
+module CP
+  class Assessment < ApplicationRecord
+    belongs_to :company
+
+    scope :latest_first, -> { order(assessment_date: :desc) }
+
+    validates_presence_of :publication_date
+
+    def emissions_all_years
```
kowal: some doc describing what is being returned here would be nice :)

app/services/import_companies.rb
```diff
@@ -39,6 +39,27 @@ def header_converter
...
+    MQ::Assessment.create!(
+      mq_assessment_attributes(row).merge(company_id: company.id)
+    )
+  end
+
+  def create_cp_assessment!(row, company)
+    assumptions = row[:assumptions]&.strip
+    cp_assessment_date = row[:carbon_performance_assessment_date]&.strip&.downcase
+
+    return if cp_assessment_date == 'not assessed' && assumptions.blank?
```
kowal: I would extract method:
```rb
def create_cp_assessment!(row, company)
  return if assessment_row_invalid?(row) # not sure about the name
  ..
```

app/services/import_companies.rb
```diff
@@ -76,6 +97,12 @@ def get_questions(row)
     end.compact
   end
 
+  def get_emissions(row)
+    row.headers.grep(/^\d{4}$/).map do |year|
```
kowal: what do you think about reformatting this like this?
```rb
    row.headers
      .grep(/^\d{4}$/)
      .map { |year| { year.to_s.to_i => row[year]&.to_f } }
      .reduce(&:merge)
      .reject { |_k, v| v.blank? }
```

tsubik: right :+1: I'm not taking much care into those importers, not sure if they will stay

kowal:
@tsubik we also need to add 'CP Assessments` to main `Assessments` tab, right? Or is it another story?

app/admin/companies.rb
tsubik: Good idea. Can try to do that with some SimpleDelegator

app/models/mq/assessment.rb
tsubik: 4STAR is like 5 level, they named it like this, we can rename it to 5, but then we will have to take care of mapping that back and forth

tsubik:
Yeah, I can add it, not sure if that's needed but it will be like 10 minutes to add.

app/services/import_companies.rb
tsubik: hmm, actually I'm striping and downcasing values and using them to create assessment later so I wouldn't move that into a separate method to not do it again.

tsubik:
Ready again @kowal

tsubik:
Also, I updated npm dependencies to make github security alert go away.

app/services/import_companies.rb
kowal: ok, that's fine then

app/models/mq/assessment.rb
kowal: so maybe just add a comment above, cause this looks like some typo ;)

PR #4, opened 2019-07-18
Feature/litigations

tsubik:
Adding litigations

```
bundle exec rake import:litigations
```


app/decorators/litigation_side_decorator.rb
```diff
@@ -0,0 +1,11 @@
+class LitigationSideDecorator < Draper::Decorator
```
kowal: this class - `LitigationSideDecorator` isn't used anywhere, is it?

tsubik: ahh, draper magic, it is used [here](https://github.com/Vizzuality/laws_and_pathways/blob/feature/litigations/app/admin/litigations.rb#L44) maybe we should do it more explicit.

PR #8, opened 2019-07-22
Feature/documents

tsubik:
The first iteration of documents upload added to Litigations

Document could be
- an uploaded one
- external (we only store URL)

TODO
- we need to add google storage service later 
- could the same document be attached to many different records? The third option "Select existing" would be needed if that is the case.
- language column is missing, will be added in another story.

upload-documents-external-links


app/models/legislation.rb
```diff
@@ -1,3 +1,18 @@
+# == Schema Information
+#
+# Table name: legislations
```
kowal: I think you need to rebase against `develop`. Not sure why github is showing those changes.

tsubik: I think that is what I did.

tsubik: or maybe not...

tsubik: done :+1:

app/decorators/litigation_decorator.rb
```diff
@@ -16,4 +16,12 @@ def summary
   def core_objective
     model.core_objective.html_safe
   end
+
+  def document_list
```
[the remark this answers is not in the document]
tsubik: good idea.

app/decorators/document_decorator.rb
```diff
@@ -0,0 +1,17 @@
+class DocumentDecorator < Draper::Decorator
+  delegate_all
+
+  def name_link
```
kowal: what do you think about renaming those methods ..
`name_link` => `document_page_link` (or `just document_link`)
`open_link` => `document_url_link`
.. to indicate where those links actually go.

app/validators/url_validator.rb
```diff
@@ -0,0 +1,16 @@
+class UrlValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return unless value.present?
```
kowal: could we combine `return unless value.present?` and `..unless url_valid?(value)` in single condition?

app/validators/url_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    record.errors[attribute] << (options[:message] || 'must be a valid URL') unless url_valid?(value)
+  end
+
+  def url_valid?(url)
+    url = begin
```
kowal: I think `url = URI.parse(url) rescue false` would be enough here

app/javascript/controllers/document_list_controller.js
```diff
@@ -0,0 +1,36 @@
+import { Controller } from 'stimulus';
+
+export default class extends Controller {
+  static targets = [ 'links', 'templateExternal', 'templateUploaded']
+
+  connect() {
+    this.wrapperClass = this.data.get('wrapperClass') || 'nested-fields';
+  }
+
+  addRecord(event) {
```
kowal: This code here looks like it shouldn't be directly related with "documents", rather, it is more like generic add/remove resource logic, right?

app/validators/url_validator.rb
tsubik: Rubocop changed that automagically.

app/javascript/controllers/document_list_controller.js
tsubik: You are right. I will change that

app/validators/url_validator.rb
tsubik: Sure, I will make two separate guard clauses. I think 2 will be more readable.

app/validators/url_validator.rb
kowal: ok, let's not mess with RuboCop 🤖

tsubik: I already did. lol. I don't like that rule

kowal: haha ok )

PR #10, opened 2019-07-26
(feature) added date_passed & document_types to Legislations

kowal:
- Added `date_passed` and `document_types` (as tags) to `Legislations`.
- extracted some import helpers

you can re-import records via `bundle exec rake import:legislation`

Zrzut ekranu 2019-07-28 o 21 39 59


app/services/import_helpers.rb
```diff
@@ -0,0 +1,17 @@
+module ImportHelpers
+  def find_location(country_iso)
+    Location.find_by!(iso: country_iso)
+  rescue StandardError
+    puts "Couldn't find Location with ISO: #{country_iso}"
+  end
+
+  def try_to_parse_date(date, expected_date_formats)
```
kowal: maybe rename it to `find_date_with_format`?

app/models/concerns/taggable.rb
```diff
@@ -1,3 +1,25 @@
+# To tag your model with let's say 'keywords', you need to
```
tsubik: Thanks for this :+1:

app/models/document_type.rb
```diff
@@ -0,0 +1,44 @@
...
+#
+#  id         :bigint           not null, primary key
+#  name       :string           not null
+#  type       :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class DocumentType < Tag
+  DOCUMENT_TYPES = [
```
tsubik: Hmm, not sure if we want to restrict that list

app/services/import_helpers.rb
```diff
@@ -0,0 +1,17 @@
+module ImportHelpers
```
tsubik: Good idea

app/services/import_helpers.rb
tsubik: You can change it, for me both are good.

app/models/document_type.rb
tsubik: Also, that list is funny now, I wouldn't pay too much attention to what is in the data currently. See there is document type `All`, they should clean this up, what they wanted to just tag Law with different types. Maybe we will have to extract some of those tags to be something else in the future (for example if application logic will depend on it, I suspect "Out of date" would be the first candidate)

app/services/import_helpers.rb
kowal: I will leave it as it is.

app/models/document_type.rb
kowal: yeah I agree, this list will definitely change. For now I can remove the validation, but eventually, I think we will need some constraint on "document type". I guess this is a more precise term than, let's say "category".

spec/controllers/admin/legislations_controller_spec.rb
```diff
@@ -2,7 +2,16 @@
 
 RSpec.describe Admin::LegislationsController, type: :controller do
   let(:admin) { create(:admin_user) }
+  let(:location) { create(:location) }
+
   let!(:legislation) { create(:legislation) }
+  let(:valid_attributes) {
+    attributes_for(:legislation).merge(
```
kowal: I couldn't find a way to include location by default in what's returned by `attributes_for` :/ then I found your approach from other models and used it here as well.

kowal:
@tsubik more updates:
- `date_passed`: added filter + default sort order + fix parsing some dates in importer
- better CSV output
- added more specs

app/models/document_type.rb
tsubik: Not sure, I think right now those are like categories, and definitely, that list needs cleanup from their side. I would just remove validation for now.

app/services/import_helpers.rb
tsubik: We can move it to `app/services/concerns`. I'm not sure if that would require to add new `concern` directory to autoload_paths in `config/application.rb`, like `config.autoload_paths += Dir[Rails.root.join("app", "services", "concerns")]`

spec/controllers/admin/legislations_controller_spec.rb
tsubik: Yes, attributes_for are not working with associations, probably they have a good reason for that.

app/services/import_helpers.rb
kowal: I think concerns are meant to encapsulate a piece of domain logic, here we have 2 simple helper methods related to different things.

Maybe I add 2 different modules, which could be used like this:

```ruby
def normalize_date
  Import::Date.try_to_parse(date)
end

def attributes
  { location: Import::Location.find(..) }
end

```

kowal: Also, I'd move all import classes to `services/import` as well, so we'll end up with `Import::Companies.new.call` and so on.

tsubik: Ok.

kowal: done, please re-review

PR #11, opened 2019-07-29
Feature/targets model

tsubik:
Adding targets, target scopes, and connecting targets with legislations

Target should be connected with sector of type "Target". We don't have types for sectors implemented yet


app/models/target.rb
```diff
@@ -0,0 +1,32 @@
...
+#  single_year      :boolean          default(FALSE), not null
+#  description      :text
+#  year             :integer
+#  base_year_period :string
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
+#
+
+class Target < ApplicationRecord
+  self.inheritance_column = nil
```
kowal: maybe to early to ask, but why do we need this here? we don't have a `type` column here.

config/initializers/active_admin.rb
```diff
@@ -272,7 +272,10 @@
   #
   #   end
 
-  config.download_links = [:csv]
+  # TODO: find other workaround, when json is not
+  # in the list above then activeadmin addon selected_list
```
kowal: can you point in a comment to activeadmin issue, if there is one?

app/models/target.rb
tsubik: I had a type column, and removed it for now. I forgot to remove this line though.

tsubik:
Where you cannot search for Legislations? When linking target with legislations? I don't have any error on my side.

PR #12, opened 2019-07-31
(feature) connect legislations & litigations

kowal:
Add ability to add **legislations** to given **litigation**

#### Litigation form:
Zrzut ekranu 2019-07-31 o 11 26 10

#### Legislation show page - added new tab (similar as for companies):
Zrzut ekranu 2019-07-31 o 14 36 13

### DIFF
--- a/app/admin/legislations.rb
+++ b/app/admin/legislations.rb
@@ -43,13 +43,10 @@ ActiveAdmin.register Legislation do
       end
 
       tab :litigations do
-        # attributes_table do
-        #   list_row 'Litigations', :litigations_links
-        # end
         panel 'Connected Litigations' do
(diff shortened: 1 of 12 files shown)


app/admin/legislations.rb
```diff
@@ -27,6 +27,38 @@
...
+          row :date_passed
+          row :location
+          row :law_id
+          row :framework
+          row :created_at
+          row :updated_at
+        end
+      end
+
+      tab :litigations do
```
tsubik: We have to think about standardizing this across the Panel. Sometimes we have it in a form of a list within a row, and sometimes as separate tabs. We can do that later when we have most of the entities in place and we will see what makes more sense.

PR #16, opened 2019-08-01
(feature)add visibility_status to Legislation and Litigation

kowal:
### Summary

- add `visibility_status` to **Legislation** and **Litigation** resources. options: `draft | pending | published | archived`, default `draft`
- I've added list scopes (above the list), so I'm not including a new filter in the sidebar
- added sidebar on show page - use can edit 'visibility_status' from this place directly

#### index page
Zrzut ekranu 2019-08-02 o 11 42 09

#### show page
- use can edit status directly from here:
Zrzut ekranu 2019-08-01 o 16 16 00

Zrzut ekranu 2019-08-01 o 16 04 13

### DIFF
--- a/app/admin/legislations.rb
+++ b/app/admin/legislations.rb
@@ -81,7 +81,6 @@ ActiveAdmin.register Legislation do
     f.inputs do
       f.input :title
       f.input :description, as: :trix
-      f.input :law_id
       f.input :document_type_ids,
               label: 'Document Types',
               as: :tags,
@@ -93,6 +92,9 @@ ActiveAdmin.register Legislation do
           f.input :framework,
                   as: :select,
                   collection: array_to_select_collection(Legislation::FRAMEWORKS)
+        end
+        column do
+          f.input :visibility_status, as: :select
         end
       end
     end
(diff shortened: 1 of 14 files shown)


db/migrate/20190801123408_add_visibility_status_to_legislation.rb
```diff
@@ -0,0 +1,5 @@
+class AddVisibilityStatusToLegislation < ActiveRecord::Migration[5.2]
+  def change
+    add_column :legislations, :visibility_status, :string
```
tsubik: maybe we can set the default value to 'draft' and not null?

tsubik: Also index on that column would be great.

app/admin/legislations.rb
```diff
@@ -24,10 +25,17 @@
     column :framework
     column :location
     column :document_types
+    tag_column :visibility_status
```
tsubik: I think we could add active admin scopes on that column like "All | Draft | Pending | Published"

kowal:
@tsubik please review again. Litigations changes will be in next PR.

PR #18, opened 2019-08-05
Feature/visibility status for companies and targets

simaob:
Similar to https://github.com/Vizzuality/laws_and_pathways/pull/17, this PR brings visibility status to Companies and Targets.


tsubik:
I think visibility status could be extracted to a `VisibilityStatus` concern later on.

PR #19, opened 2019-08-06
Feature/as google cloud storage

tsubik:
Configuring GCS for ActiveStorage file upload.

~On staging/prod we use credentials encrypted file, but `GCS_CREDENTIALS` and bucket name could be set using ENV variables on development env.~

`GCS_CREDENTIALS_FILE` to override file name with credentials.


tsubik:
Tests fixed, and strategy to keep credentials json file changed. Those should be stored in `config/secrets` directory and properly named. Capistrano is storing credentials in `shared` folder and making a symbolic link for a specific file.

PR #21, opened 2019-08-06
Feature/register created by and updated by

simaob:
This Pull Request adds:

- UserTrackable and UserTracker concerns to enable keeping track of who added a record and who edited it last;
- It integrates the above with litigations;
- It updates litigations admin pages with the created_by and updated_by;
- Updates litigations controller specs to handle this. Had to manually delete the current_user key, as it was persisting between specs and causing non existing users to be used in new specs. Wondering if this is the best way to do it. But I spent 2 hours banging my head against it, so I'll go with this for now.


app/controllers/application_controller.rb
```diff
@@ -1,2 +1,9 @@
 class ApplicationController < ActionController::Base
+  before_action :set_current_user
+
+  private
+
+  def set_current_user
+    AdminUser.current_admin_user = current_admin_user
```
tsubik: I saw I think something new in Rails 5.2 https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html
From what I read this does exactly what you need, but without an extra gem. Maybe it's worth a try?

simaob: Alright, I haven't heard about this =D I'll take a look at it tomorrow! =)

simaob: This is exactly what we need. =) I've updated the code to use this instead of the request_store. I seem to still need to clear the Current.admin_user on the before(:each) of the litigations controller spec, otherwise my specs fail :/

Basically I'd expect the before(:each) to always sign in the new admin_user, but it seems that we need clear it by hand or maybe we could log it out. Any thoughts?

Cheers!

app/models/concerns/user_tracker.rb
```diff
@@ -0,0 +1,13 @@
+module UserTracker
```
tsubik: Is this needed now?

app/controllers/application_controller.rb
tsubik: I will look at the specs.

app/models/concerns/user_tracker.rb
simaob: good question. probably not =D let me check.

simaob: specs pass without it, so we probably don't need it =)

simaob: removed

app/controllers/application_controller.rb
tsubik: Crap, I have no idea why is that not removed, but maybe we can add global clear of all current attributes after each controller test?

```
  config.after :each, type: :controller do
    ActiveSupport::CurrentAttributes.reset_all
  end
```

tsubik: in rails_helper

tsubik: I think controller specs are not just invoking the actual reset code so that could be a good workaround. Maybe it's worth to look at request specs instead of controller.

simaob: oh right, that might be it. I'll add your thing to the rails_helper for the moment. Spent too much time on this already :p

tsubik:
Forgot to say that is also worth to add updated_by, created_by to view page to be like updated_at, updated_by, created_at, created_by

simaob:
you mean the show page? I added it to the index.

simaob:
or you mean including the dates as well?

tsubik:
Show page.

tsubik:
I think it wasn't there.

tsubik:
And we have annotate gem, so you can `be annotate`

simaob:
Done! I've left the dates out of the index table for now, as I'm not sure if it isn't too much.

tsubik:
Yeah, dates should not be on index page, I agree.

PR #22, opened 2019-08-07
Feature/revisit core entities

tsubik:
Couple models updated

- Adding Keywords for Legislation and Litigation model, common tag system
- Adding Natural Hazards for Legislation
- Allow multiple values for Legislation framework, make Framework model manageable by admin
- Add INDC URL to Locations

Keywords and Natural Hazards are more open tags, which means a user can type and dynamically add new entries directly in the input.

### DIFF
--- a/app/admin/frameworks.rb
+++ b/app/admin/frameworks.rb
@@ -1,0 +1,9 @@
+ActiveAdmin.register Framework do
+  config.sort_order = 'name_asc'
+
+  menu parent: 'Tags'
+
+  permit_params :name
+
+  filter :name_contains, label: 'Name'
+end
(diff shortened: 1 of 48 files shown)


tsubik:
Forgot to mention would be great to do `bundle exec import:all`

tsubik:
Those added in bulk will be added in input tag anyway, but I think that's a good idea to simplify Tag pages. Not sure if it is also worth removing `show` page, not much there. Probably `edit` page should be like a default one.

tsubik:
I will do that in separate PR.

simaob:
Makes sense, we don't need a show page for these type of resources! Cheers!

PR #23, opened 2019-08-07
Feature/propagate created by and updated by

simaob:
Adds UserTrackable to:

- legislations;
- locations;
- targets;

It also fixes the issue with the ApplicationController not being found in the tree, by adding `::` to `Current`, following this thread: https://stackoverflow.com/a/18423438/2701824

It removes some decorators as it turns out they are not needed!


tsubik:
And that error looks like is fixed, nice!

PR #25, opened 2019-08-08
better structure for admin javascript files, quick gravatar

tsubik:
Maybe not a priority but nice to have and I had to do a little cleanup in js files structure anyway.

image


tsubik:
Yeah, both issues come from ActiveAdmin and that template. I will take a look at the long email one here.

tsubik:
@simaob fixed

PR #27, opened 2019-08-08
Moved shared published functionality to concern

simaob:
This PR extracts the visibility status usage to a concern to be shared by different models.

I haven't done the same to the Admin pages under `app/admin` as I wasn't able to find a lot of guidance. I thought about extracting the scopes into a new concern for the admin bits, but not sure if it's worth it.

One example I found on Stack Overflow mentions using `extend` instead of include, because it seems that Active Admin doesn't support the ActiveSupport::Concern: https://stackoverflow.com/questions/27325605/dry-ing-up-activeadmin

Also found this word of warning: https://tmichel.github.io/2015/02/22/sharing-code-between-activeadmin-resources/

### DIFF
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -2,11 +2,7 @@ ActiveAdmin.register Company do
   menu priority: 0, parent: 'TPI'
   config.sort_order = 'name_asc'
 
-  scope('All', &:all)
-  scope('Draft', &:draft)
-  scope('Pending', &:pending)
-  scope('Published', &:published)
-  scope('Archived', &:archived)
+  publishable_scopes
 
   permit_params :name, :isin, :sector_id, :location_id, :headquarter_location_id,
                 :ca100, :size, :visibility_status
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -3,6 +3,7 @@ ActiveAdmin.register Company do
   config.sort_order = 'name_asc'
 
   publishable_scopes
+  publishable_sidebar only: :show
 
   permit_params :name, :isin, :sector_id, :location_id, :headquarter_location_id,
                 :ca100, :size, :visibility_status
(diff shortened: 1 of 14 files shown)


tsubik:
Yeah, I know doing many things in AA is hard. Extending DSL is the best option here. There are some AA extensions already in the code I have a little cleanup in mind, maybe I could try to add that extension here.

simaob:
Do it! :D

tsubik:
I'm also thinking if scope `all` should return `archived` or maybe the name all is not great here. Creating default scope without `archived`? not sure what other problems we will have after that.

simaob:
We could not have "All" and just keep the specific ones, for clarity. Not sure when they'd need them all in a page.

simaob:
Maybe default to Published.

tsubik:
I'm talking about model `default_scope`

tsubik:
That archived sounds for me like "soft-deleted". That should not be present in any select where you can link association.

tsubik:
We need `All` like this is default scope that you get when you enter view, if we remove it then the only way to go back to see all would be page refresh.

simaob:
oh right, yes, I get it. We should prevent archived objects from being linked to other objects, as they are archived. Makes sense to exclude archived from that.

tsubik:
Exactly, take a look at last commit, not sure what else could go into publishable "concern" for ActiveAdmin. I think loading extensions in this way is cleaner (to have them separately and load in initializer), what do you think?

simaob:
That's quite nice and tidy. when I was looking at it, I couldn't see much else either. Potentially the sidebar bit for the show page?

tsubik:
Right, I thought that only one sidebar block is allowed but apparently we can add as many separate sidebar blocks as we want. Good idea, I will add it as a separate `publishable_sidebar` method.

simaob:
Thanks for the help! =)

PR #28, opened 2019-08-09
Renames locations model to geographies

simaob:
- updates all references of locations to use the term geographies;
- updates indeces, foreign_keys and column attributes;
- updates models, active admin register files, tasks, specs and views;
- fixes typo in `headquarter_geography_id`, it now reads `headquarters`.


tsubik:
And also rerun model annotations.

tsubik:
Attributes names scope needs to be renamed in `config/locales/en.yml`

simaob:
Thanks for spotting those, I clearly missed a few places! Hope I tackled them all now!

PR #29, opened 2019-08-11
Feature/add language input to documents

kowal:
#### Summary

Add language input/label to
- document show page
- documents index page
- document form (in the litigation context)

Languages are taken from https://github.com/scsmith/language_list.

Zrzut ekranu 2019-08-11 o 17 37 21

Zrzut ekranu 2019-08-12 o 14 58 18

### DIFF
--- a/app/decorators/litigation_decorator.rb
+++ b/app/decorators/litigation_decorator.rb
@@ -21,10 +21,10 @@ class LitigationDecorator < Draper::Decorator
     return [] if model.documents.empty?
 
     model.documents.map do |document|
-      language = LanguageList::LanguageInfo.find(document.language).name
-      link_text = "#{document.name} (language: #{language}, #{document.external? ? 'external' : 'uploaded'})"
-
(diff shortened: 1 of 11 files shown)


app/decorators/litigation_decorator.rb
```diff
@@ -21,7 +21,10 @@ def document_links
     return [] if model.documents.empty?
 
     model.documents.map do |document|
-      h.link_to document.name, document.url, target: '_blank'
+      h.link_to document.name,
+                document.url,
+                target: '_blank',
+                title: document.external? ? document.url : nil
```
tsubik: not sure why the title should be `document.url` here and not always `document.name`

kowal: I guess it could be handy for users to see where those external links are pointing.

kowal:
> I think you can add `language` to factory too

it was already there, but I'll add comment on what format to use

app/decorators/litigation_decorator.rb
tsubik: ahh, ok

PR #30, opened 2019-08-12
add cancel button on Litigation form

kowal:
Zrzut ekranu 2019-08-12 o 15 52 55

### DIFF
--- a/app/views/admin/litigations/_form.html.erb
+++ b/app/views/admin/litigations/_form.html.erb
@@ -41,5 +41,8 @@
     </div>
   </div>
 
-  <%= f.actions %>
+  <%= f.actions do %>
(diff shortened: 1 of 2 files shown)


app/views/admin/litigations/_form.html.erb
```diff
@@ -41,5 +41,8 @@
     </div>
   </div>
 
-  <%= f.actions %>
+  <%= f.actions do %>
```
tsubik: so this is weird, arbre `f.actions` when in creating form in controller directly has different output than in custom forms.

kowal: Yep, I was trying to find an upstream issue, no luck. Maybe I'll file a bug later on.

PR #31, opened 2019-08-12
Feature/enable multiple isins

simaob:
This small PR allows the input of multiple isin codes for the same company.

- render as strings;
- using a tags input to manage them;
- updating the importer to allow saving multiple isins, matching the import on the company name.

### DIFF
--- a/app/models/company.rb
+++ b/app/models/company.rb
@@ -48,8 +48,4 @@ class Company < ApplicationRecord
   def to_s
     name
   end
-
-  def isin_string
-    isin.split(',')
-  end
 end
--- a/app/models/company.rb
+++ b/app/models/company.rb
@@ -48,4 +48,8 @@ class Company < ApplicationRecord
   def to_s
     name
   end
+
+  def isin_string
+    isin.split(',')
+  end
 end
--- a/app/services/import/companies.rb
+++ b/app/services/import/companies.rb
@@ -19,8 +19,11 @@ module Import
       import_each_with_logging(csv, FILEPATH) do |row|
         company = Company.find_or_initialize_by(name: row[:company_name])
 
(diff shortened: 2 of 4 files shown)


app/services/import/companies.rb
```diff
@@ -17,7 +17,14 @@ def call
 
     def import
       import_each_with_logging(csv, FILEPATH) do |row|
-        company = Company.find_or_initialize_by(isin: row[:isin])
+        company = Company.find_or_initialize_by(name: row[:company_name])
+
+        company.isin = if company.isin.present?
```
kowal: couldn't this be moved to `company_attributes` method?

simaob: probably so.  I kept it here, but it doesn't make so much sense now. I'll move it.

app/models/company.rb
```diff
@@ -48,4 +48,8 @@ def latest_assessment
   def to_s
     name
   end
+
+  def isin_string
```
kowal: I think this belongs to `CompanyDecorator` which needs to be created :)

kowal: also, this name is a bit confusing - this returns an array.

app/services/import/companies.rb
simaob: hmmm, actually, I would need to pass the current isin to that method if I were to manage this there. Currently the `company_attributes` is not aware of the record being updated. So not sure if that's worth it.

app/models/company.rb
simaob: you're right! I've moved the method to a new decorator, and I've renamed it to `isin_as_tags`. 🙇

kowal: 👍

PR #32, opened 2019-08-13
Feature/add documents to legislations

kowal:
#### Summary
- update model relations
- update controller to allow saving documents within legislations
- update controller specs to check if all data was properly saved (legislations, litigations)
- update importer to create documents from `source_text_links` column

UI form changes will be in another PR.

after `be rake import:legislation`:
Zrzut ekranu 2019-08-14 o 13 20 38

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -31,6 +31,8 @@ class Document < ApplicationRecord
 
   validates_presence_of :name, :type
 
+  scope :from_documentable, ->(documentable) { where(documentable_type: documentable) }
+
   def url
     return file_url if uploaded?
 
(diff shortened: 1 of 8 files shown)


app/models/document.rb
```diff
@@ -31,16 +31,14 @@ class Document < ApplicationRecord
...
 
+  scope :from_documentable, ->(documentable) { where(documentable_type: documentable) }
+
   def url
     return file_url if uploaded?
 
     external_url
   end
 
-  def external?
```
kowal: this was redundant

app/models/document.rb
```diff
@@ -50,6 +48,6 @@ def file_url
   end
 
   def set_last_verified_on
-    self.last_verified_on = Time.zone.now.to_date
+    self.last_verified_on ||= Time.zone.now.to_date
```
kowal: importer needs to set custom verification date

PR #34, opened 2019-08-18
add document upload to UI for legislations

tsubik:
There are some problems with using both `arbre` and `erb` templates like something is not rendering or rendering twice. Maybe we should stick with erb in most cases, just need to find a better alternative for styling columns, for now I used `Arbre::Context` and column component.


app/views/admin/legislations/_form.html.erb
```diff
@@ -0,0 +1,49 @@
...
+
+    <div class="tab-content">
+      <div id="details">
+        <%= f.inputs do %>
+          <%= f.input :title %>
+          <%= f.input :description, as: :trix %>
+          <%= f.input :document_type_ids, label: 'Document', as: :tags, collection: DocumentType.all %>
+          <%=
+            Arbre::Context.new do
+              columns do
```
kowal: @tsubik So I've added column layout here without specific requirement, I thought the form will be more compact and easy to scan this way. I guess we can drop columns from here for now until we get more specific details about form UI guidelines.

app/decorators/legislation_decorator.rb
```diff
@@ -25,4 +29,15 @@ def litigations_links
                 title: litigation.title
     end
   end
+
+  def document_links
```
kowal: we could extract this to `Decorators::FromDocuments` to share common `Document` - related presentation logic.

PR #35, opened 2019-08-20
Fix selected list positioning bug

tsubik:
Pasted Image at 2019_08_20_11_01 pm


kowal:
For reference - this was fixed in select 4.0.9  in `Fix dropdown positioning when displayed above with messages (#5592))`, however rails wrapper is still on version 4.0.3 :(. We could create our own fork of `select2-rails` with 4.0.9 bundled and use it. Since it can break other things let's consider it as a separate technical task.

PR #38, opened 2019-08-22
Feature/external legislation

kowal:
Added `ExternalLegislation` resource + CRUD

Zrzut ekranu 2019-08-23 o 12 34 48
Zrzut ekranu 2019-08-23 o 12 34 56

### DIFF
--- a/app/admin/documents.rb
+++ b/app/admin/documents.rb
@@ -1,4 +1,6 @@
 ActiveAdmin.register Document do
+  config.batch_actions = false
+
   menu priority: 6
 
   decorate_with DocumentDecorator
@@ -8,8 +10,6 @@ ActiveAdmin.register Document do
   filter :language,
          as: :select,
          collection: proc { all_languages_to_select_collection }
-
-  config.batch_actions = false
 
   actions :all, except: [:new, :edit, :create, :update]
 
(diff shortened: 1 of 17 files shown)


db/migrate/20190822112921_create_external_legislations.rb
```diff
@@ -0,0 +1,11 @@
+class CreateExternalLegislations < ActiveRecord::Migration[5.2]
+  def change
+    create_table :external_legislations do |t|
+      t.string :name
```
tsubik: we could add not null here

kowal:
> 👍 Just one small change to add not null and validation.
> 
> Adding external legislation to litigation form is a separate task? I don't remember. We could use selectedList component the same as for normal legislations.

yes, it will be separate task/PR

db/migrate/20190822112921_create_external_legislations.rb
kowal: updaded

app/admin/documents.rb
```diff
@@ -1,4 +1,6 @@
 ActiveAdmin.register Document do
+  config.batch_actions = false
```
kowal: I moved `config.` statements to the top everywhere

tsubik:
Ahh ok, that was as another task. I will merge this one then.

PR #41, opened 2019-08-26
Chore/upload import cleanup

kowal:
#### Summary

- rename `BaseUploader` to `BaseImporter`
- rename & unify `details/stats` to `upload_results`
- rename `import_each_with_logging` to `import_each_csv_row `
- do not call `parse_csv` twice during import

### DIFF
--- a/app/services/csv_import/base_importer.rb
+++ b/app/services/csv_import/base_importer.rb
@@ -1,17 +1,25 @@
-module Upload
-  class BaseUploader
+module CSVImport
+  class BaseImporter
     include ActiveModel::Model
 
-    attr_reader :file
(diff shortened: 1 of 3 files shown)


app/services/csv_import/base_importer.rb
```diff
@@ -1,17 +1,23 @@
...
+module CSVImport
+  class BaseImporter
     include ActiveModel::Model
 
-    attr_reader :file
+    attr_reader :file, :import_results
 
     def initialize(file)
       @file = file
+      @import_results ||= {
```
tsubik: here could be only `=`

kowal: fixed

PR #43, opened 2019-08-28
Feature/documents export

kowal:
### Summary

* Add a link to `documents.csv` on the **legislations** and **litigations** index pages.
* update CSV columns for documents to `Id, Name, External URL, language, Last verified on, Law ID, Documentable ID, Documentable Type`
* added new controller specs for CSV download

Zrzut ekranu 2019-08-29 o 15 10 15

--------

Zrzut ekranu 2019-08-29 o 15 10 54

### DIFF
--- a/app/admin/documents.rb
+++ b/app/admin/documents.rb
@@ -37,19 +37,18 @@ ActiveAdmin.register Document do
 
   controller do
     def scoped_collection
-      return super.includes(:documentable) unless params[:q].present?
+      results = super.includes(:documentable)
 
-      documentable_type = params[:q][:documentable_type_eq]
-      documentable_query_params = params[:q][:documentable]
-      return unless documentable_type.present? && documentable_query_params.present?
+      documentable_type = params.dig(:q, :documentable_type_eq)
+      documentable_query_params = params.dig(:q, :documentable)
+      if documentable_query_params.present? && documentable_query_params.present?
+        documentable_klass = documentable_type.constantize
+        results = results.where(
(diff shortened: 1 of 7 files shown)


app/admin/documents.rb
```diff
@@ -37,7 +37,32 @@
...
   controller do
     def scoped_collection
-      super.includes(:documentable)
+      results = super.includes(:documentable)
+
+      documentable_klass = find_documentable_klass
+
+      if documentable_klass.present? && documentable_params.present?
+        results = results.where(
+          documentable: documentable_klass.ransack(documentable_params).result
```
kowal: @tsubik thanks for improving this! works like a charm now)

app/admin/documents.rb
```diff
@@ -35,9 +35,45 @@
     actions
   end
 
+  csv do
+    column :id
+    column :name
+    column :external_url
+    column :language
+    column :last_verified_on
+    column 'Law ID', &:document_legacy_law_id
```
tsubik: I'm not sure if we need that legacy law ID for our import/export.

kowal: This is the only way to match documents with legislations. This could be handy during: `export -> manual change -> import` scenario. Also, it doesn't break anything.

I do agree that in the future, after the initial import, this probably will be removed.

tsubik: That's documentable_id for.

kowal: ok, I removed this for now

PR #46, opened 2019-08-28
Feature/account editing flow

tsubik:
- Adding first name and last name to admin users
- Changing password is in the different tab now, and you can change other attributes without changing password (not sure why that was by default)
- Adding gravatar to admin forms and general clean up

Related story: https://www.pivotaltracker.com/story/show/167780870


tsubik:
admin-users

PR #48, opened 2019-09-01
download events CSV for legislation and litigations

kowal:
### Events CSV download

- Added basic active admin's `events` resource, the whole index pages isn't required (I think) however I needed index controller action (with proper routes). Not sure if I can keep only this.

Zrzut ekranu 2019-09-01 o 15 29 53
Zrzut ekranu 2019-09-01 o 21 08 25
Zrzut ekranu 2019-09-01 o 21 09 35

### DIFF
--- a/app/admin/events.rb
+++ b/app/admin/events.rb
@@ -22,24 +22,19 @@ ActiveAdmin.register Event do
     def scoped_collection
       results = super.includes(:eventable)
 
-      eventable_klass = find_eventable_klass
+      return results unless eventable_params
 
-      if eventable_klass.present? && eventable_params.present?
-        results = results.where(
-          eventable: eventable_klass.ransack(eventable_params).result
-        )
-      end
-
-      results
+      results.where(eventable: eventable_scope)
     end
 
     private
 
-    def find_eventable_klass
-      eventable_type = params.dig(:q, :eventable_type_eq)
-      return nil unless eventable_type
+    def eventable_scope
+      find_eventable_klass&.ransack(eventable_params)&.result
+    end
 
-      eventable_type.constantize
+    def find_eventable_klass
+      @eventable_klass ||= params.dig(:q, :eventable_type_eq)&.constantize
     rescue NameError => e
       raise "Can't find eventable class based on given 'eventable_type_eq' param: #{e.message}"
     end
--- a/app/admin/events.rb
+++ b/app/admin/events.rb
@@ -1,0 +1,51 @@
+ActiveAdmin.register Event do
+  menu label: 'All Events', priority: 3, parent: 'Administration'
+
+  actions :all, except: [:update, :show, :destroy]
+
+  INDEX_COLUMNS_DEFINITION = proc do
+    column :id
+    column :eventable_type
+    column :eventable_id
+    column :event_type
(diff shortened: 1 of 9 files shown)


app/admin/events.rb
```diff
@@ -0,0 +1,51 @@
+ActiveAdmin.register Event do
+  menu label: 'All Events', priority: 3, parent: 'Administration'
```
kowal: question: how to keep only `index` action for CSV download, but without adding the 'events' page anywhere?

app/admin/events.rb
```diff
@@ -0,0 +1,51 @@
...
+          eventable: eventable_klass.ransack(eventable_params).result
+        )
+      end
+
+      results
+    end
+
+    private
+
+    def find_eventable_klass
```
kowal: this could be extracted to smth common, together with `find_documentable_klass`

config/initializers/active_admin.rb
```diff
@@ -41,16 +41,16 @@
...
-      admin.build_menu do |menu|
-        menu.add label: 'Geographies', priority: 1
-        menu.add label: 'Laws', priority: 2
-        menu.add label: 'TPI', priority: 3
-        menu.add label: 'Tags', priority: 7
-        menu.add label: 'Administration', priority: 8
-      end
+  config.namespace :admin do |admin|
+    # admin.site_title = "Custom Admin Title"
+    admin.build_menu do |menu|
```
kowal: just formatting

app/admin/events.rb
tsubik: You can add `menu: false`

app/admin/events.rb
```diff
@@ -0,0 +1,46 @@
+ActiveAdmin.register Event do
+  menu label: 'All Events', priority: 3, parent: 'Administration'
+
+  actions :all, except: [:new, :edit, :update, :show, :destroy]
```
tsubik: could be also `actions :index` if only one action.

app/admin/events.rb
```diff
@@ -0,0 +1,46 @@
...
+    column :date
+    column :url
+  end
+
+  index { instance_exec(&INDEX_COLUMNS_DEFINITION) }
+
+  csv { instance_exec(&INDEX_COLUMNS_DEFINITION) }
+
+  controller do
+    def scoped_collection
```
tsubik: to redirect to for example admin dashboard instead of displaying index page, you could do like

```
    def index
      super do |format|
        format.html { redirect_to admin_dashboard_path }
      end
    end
```
That would preserve format.csv response

app/admin/events.rb
```diff
@@ -0,0 +1,46 @@
...
+    column :eventable_type
+    column :eventable_id
+    column :event_type
+    column :title
+    column :description
+    column :date
+    column :url
+  end
+
+  index { instance_exec(&INDEX_COLUMNS_DEFINITION) }
```
tsubik: You could remove it if we don't need the page with all events

tsubik: I think we don't need that page

kowal:
@tsubik all fixed

PR #51, opened 2019-09-04
Feature/data upload legislations

kowal:
#### Summary

* added uploader handler for Legislations
* renamed `DataUploadForm` to `Command::CsvDataUpload`
  - it is more like use case object than related with form
  - I wrapped it in `Command` namespace - I think we'll have more domain commands in the future
* added tests for `Command::CsvDataUpload`
* extracted `find_record_by` and `resource_klass` methods into `BaseImporter`

Zrzut ekranu 2019-09-04 o 22 27 49
Zrzut ekranu 2019-09-04 o 22 28 09


spec/commands/csv_data_upload_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'rails_helper'
+
+describe 'CsvDataUpload (integration)' do
+  let(:legislations_simple_csv) { fixture_file('legislations_simple.csv') }
+  let(:litigations_simple_csv) { fixture_file('litigations_simple.csv') }
+
+  let!(:countries) do
+    [
+      create(:geography, iso: 'PL'),
```
tsubik: let's use even in tests iso3 codes.

spec/commands/csv_data_upload_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'rails_helper'
+
+describe 'CsvDataUpload (integration)' do
```
tsubik: :+1: for tests

config/initializers/active_admin.rb
```diff
@@ -1,6 +1,9 @@
 Dir[Rails.root.join('lib/extensions/active_admin/*.rb')].each { |f| require f }
 
 ActiveAdmin.setup do |config|
+  # Custom Load paths
+  config.load_paths += [File.join(Rails.root, 'app', 'commands')]
```
tsubik: Hmm, and other not standard paths like `services` should not be added here as well? Is that needed?

kowal: you're right! I just had to move the file to `command` directory

PR #53, opened 2019-09-06
make tabs URL addressable

kowal:
Tabs on all 'show' pages are now URL-addressable, which means switching tabs => updating URL via stimulus `tabs_controller`.

Please note that we still have a problem that both tabs and main headers have same `#id`, so for bigger site content, page scrolls down to tab **content** header, which can be annoying sometimes.

### DIFF
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -65,15 +65,9 @@ ActiveAdmin.register Company do
                 end
 
                 table_for a.questions do
-                  column :level do |q|
-                    q['level']
-                  end
-                  column :answer do |q|
-                    q['answer']
-                  end
-                  column :question do |q|
-                    q['question']
-                  end
+                  column(:level) { |q| q['level'] }
+                  column(:answer) { |q| q['answer'] }
+                  column(:question) { |q| q['question'] }
                 end
               end
             end
@@ -120,17 +114,14 @@ ActiveAdmin.register Company do
   end
 
   index do
-    column :name do |company|
-      link_to company.name, admin_company_path(company)
-    end
+    column(:name) { |company| link_to company.name, admin_company_path(company) }
     column :isin, &:isin_as_tags
-    column :size do |company|
-      company.size.humanize
-    end
+    column(:size) { |company| company.size.humanize }
     column :level, &:mq_status_description_short
     column :geography
     column :headquarters_geography
     tag_column :visibility_status
+
     actions
   end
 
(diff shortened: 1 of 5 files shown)


app/admin/companies.rb
```diff
@@ -65,15 +65,9 @@
...
-                  column :level do |q|
-                    q['level']
-                  end
-                  column :answer do |q|
-                    q['answer']
-                  end
-                  column :question do |q|
-                    q['question']
-                  end
+                  column(:level) { |q| q['level'] }
```
kowal: just formatting

app/admin/companies.rb
```diff
@@ -120,17 +114,14 @@
   end
 
   index do
-    column :name do |company|
-      link_to company.name, admin_company_path(company)
-    end
+    column(:name) { |company| link_to company.name, admin_company_path(company) }
```
kowal: just formatting

lib/extensions/active_admin/active_admin_addressable_tabs.rb
```diff
@@ -0,0 +1,8 @@
+module ActiveAdminAddressableTabs
+  # Creates tabs component which enables stimulus tabs_controller
+  def tabs(&block)
+    div 'data-controller': 'tabs' do
+      super(&block)
```
kowal: Unfortunately, we can't add custom HTML on `tabs` level here, we can only wrap it in custom `div`.

simaob:
looking at this now! =)

simaob:
Hey Kowal, I'm having trouble seeing this and not sure if it's my local setup that is broken. When I run the app with `rails s` I'm not seeing any anchors being set. When I run it with `yarn start`, I'm getting some time out and really slow loads. Any thoughts?

kowal:
@simaob I checked this again - works fine for me. Maybe try restarting spring (`./bin/spring stop` and `yarn start` again). Spring is causing some code-cashing issues for me from time to time.

app/admin/documents.rb
```diff
@@ -29,7 +29,8 @@
 
   index do
     column 'Name', &:document_page_link
-    column 'Attached To', :documentable
+    column 'Resource Type', :documentable_type
```
tsubik: Not sure if that would be more descriptive.

tsubik:
I found a way to remove that scroll to content when changing the tab. Instead of changing window hash property in tabs controller, we can `history.replaceState({}, '', tabName);`.

PR #54, opened 2019-09-06
Feature/bulk upload companies

kowal:
* adds bulk importer for `companies`

### DIFF
--- a/app/services/csv_import/companies.rb
+++ b/app/services/csv_import/companies.rb
@@ -39,13 +39,20 @@ module CSVImport
       {
         name: row[:name],
         isin: row[:isin],
-        sector: Sector.find_or_create_by!(name: row[:sector]),
+        sector: find_or_create_sector(row),
         size: row[:size],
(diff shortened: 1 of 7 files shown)


app/services/csv_import/companies.rb
```diff
@@ -0,0 +1,51 @@
...
+    # Builds resource attributes from a CSV row.
+    #
+    # If not present in DB, will create related:
+    # - sector
+    #
+    def company_attributes(row)
+      {
+        name: row[:name],
+        isin: row[:isin],
+        sector: Sector.find_or_create_by!(name: row[:sector]),
```
tsubik: I would try to find by downcased name, to prevent creating multiple sectors with the same name.

```
    def find_or_create_sector(row)
      return unless row[:sector].present?

      Sector.where('lower(name) = ?', row[:sector].downcase).first ||
        Sector.new(name: row[:sector])
    end
```

kowal: updated

PR #55, opened 2019-09-10
[IWP] Feature/archive soft delete

kowal:


app/admin/legislations.rb
```diff
@@ -109,9 +109,35 @@
...
+               if: proc { current_scope&.name != 'Archived' } do |ids|
+    archive_command = ::Command::Batch::Archive.new(batch_action_collection, ids)
+
+    results = if archive_command.call
+                {notice: "Successfully archived #{ids.count} Legislations"}
+              else
+                {alert: 'Could not archive selected Legislations'}
+              end
+
+    redirect_to collection_path(scope: 'archived'), results
```
kowal: set scope to `archive` after batch-archiving records

app/commands/command/destroy/legislation.rb
```diff
@@ -0,0 +1,24 @@
+module Command
+  module Destroy
+    class Legislation
```
kowal: This command defines what it means to destroy Legislation.

TODO: need to add tests for this

app/commands/command/batch/archive.rb
```diff
@@ -0,0 +1,16 @@
+module Command
+  module Batch
+    class Archive
+      def initialize(collection, ids)
+        @collection = collection
+        @ids = ids
+      end
+
+      def call
+        @collection
```
kowal: Do we want to perform some extra checks before (batch)-archiving? Are there any cases when records (legislations in particular) should not have option to be archived? Those could be modeled using AM validations, similar as in `CsvDataUpload`

app/models/legislation.rb
```diff
@@ -19,9 +19,12 @@
 class Legislation < ApplicationRecord
   include UserTrackable
   include Taggable
-  include Publishable
+  include VisibilityStatus
+  include Discard::Model
   extend FriendlyId
 
+  default_scope -> { kept } # hide deleted records
```
kowal: `kept` is a scope provided by `discard` gem

simaob:
Good job so far!

Comment: the deleted success message should probably say "deleted" instead of "destroyed", just for language sake.

Answer to:

>deleting associated documents and events (⚠️ @simaob do we really want this?)

I'm wondering if we should discard/soft delete events, but delete documents (as they would take space in the Google Cloud Storage). What do you think? We can also confirm with them and potentially just keep the documents, if it's not too expensive.

But overall this is the way!

simaob:
Having discussed this over slack we decided to go with discarding documents and events.

app/commands/command/batch/archive.rb
simaob: I'm thinking that this will be safeguarded by the authorization rules, where not everyone will have access to all features. (I know, I need to finish that bit!). So for now, I wouldn't spend too much time thinking. As it will be brought in or handled from a different side.

kowal:
> Comment: the deleted success message should probably say "deleted" instead of "destroyed", just for language sake.

it is already :)

kowal:
I also need to overwrite batch-delete action - default one still destroys objects 😅

PR #56, opened 2019-09-12
(feature): bulk upload for Targets

kowal:
Zrzut ekranu 2019-09-12 o 18 03 25

Zrzut ekranu 2019-09-13 o 09 09 51

Zrzut ekranu 2019-09-12 o 18 03 41


app/services/csv_import/companies.rb
```diff
@@ -37,6 +37,7 @@ def prepare_company(row)
     #
     def company_attributes(row)
       {
+        id: row[:id],
```
kowal: I think we might include this mapping `id: row[:id]` in all csv importers.

Please consider the following scenario:

1. User exports CSV (`id` column are present in the file)
2. The user deletes one of the records in the application (it is actually soft-deleted)
3. The user then re-imports CSV with `id` column - he would expect that those ids from the file are going to be used and effectively he'd update deleted record, not create a new one.

Without the mapping, we'd re-create the same record but with different id - I guess we don't want that, we want that CSV import would preserve discarded records just as they are, right?

kowal: @simaob @tsubik what do you think?

spec/commands/csv_data_upload_spec.rb
```diff
@@ -1,9 +1,10 @@
 require 'rails_helper'
 
 describe 'CsvDataUpload (integration)' do
-  let(:legislations_simple_csv) { fixture_file('legislations_simple.csv') }
-  let(:litigations_simple_csv) { fixture_file('litigations_simple.csv') }
-  let(:companies_simple_csv) { fixture_file('companies_simple.csv') }
+  let(:legislations_csv) { fixture_file('legislations.csv') }
```
kowal: as @tsubik suggested some time ago, I removed `_simple` suffix as it doesn't bring much value.

spec/factories/targets.rb
```diff
@@ -32,6 +32,6 @@
     ghg_target { false }
     single_year { true }
     target_type { 'base_year_target' }
-    visibility_status { Litigation::VISIBILITY.sample }
+    visibility_status { Litigation::VISIBILITY.first }
```
kowal: let's remove this randomness from specs ;)

app/services/csv_import/companies.rb
kowal: removed this for now, let's move the discussion to another story.

PR #57, opened 2019-09-13
(feature): TPI (sectors, companies) controllers, chart API & views

kowal:
## Summary

Used https://chartkick.com/#highcharts with **highcharts**.

### Sectors index
- [x] Management Quality: All sectors (pie chart)
  - [x] ~Market Cap (bubble chart)~ 🤕 not in this PR
- [x] Carbon Performance: All sectors (stacked chart)

### Sector Details page
- [x] Management Quality: Sector (pie chart)
  - [x] Levels (NOT chart, just flexbox)
- [x] Carbon Performance: Sector (line chart)

### Company Details page
- [x] ~Management Quality: Company (gauge?)~ not in this PR
- [x] ~Management Quality: Company (line chart)~  not in this PR
- [x] Company question/answers list
- [x] Carbon Performance: Company (line chart)
- [x] Company Info

------
TODO:
- [x] fix issues with assessment date
- [x] use real data for `Carbon Performance: All sectors` chart (@agwozdowski)
- [x] add controllers or integration specs
------
Go to http://localhost:3000/tpi/sectors
Zrzut ekranu 2019-09-14 o 08 33 14


app/views/sectors/index.html.erb
```diff
@@ -0,0 +1,51 @@
...
+<%= @companies_names.map {|c| link_to c.last, company_path(c.first)}.join(', ').html_safe %>
+<br>
+<br>
+
+<h3>Management Quality: All sectors</h3>
+<p>Distribution of companies in All sectors according to the management of their greenhouse gas emissions and of risks and opportunities to the low-carbon transition.</p>
+
+<div class="chart">
+<%= pie_chart(
+      companies_levels_sectors_path,
```
kowal: instead of data, we provide URL so it gets fetched after the main page is rendered.

app/services/api/sectors/companies_count_grouped_by_scenario.rb
```diff
@@ -0,0 +1,46 @@
...
+      #   name: 'Below 2',
+      #   data: [ ['Coal Mining', 52], ['Steel', 73] ]
+      #  },
+      #  {
+      #    name: 'Paris',
+      #    data: [ ['Coal Mining', 65], ['Steel', 26] ]
+      #  }
+      # ]
+      def get
+        CP_SCENARIOS.map do |cp_alignment|
```
kowal: this needs to be rewritten after https://github.com/Vizzuality/laws_and_pathways/pull/59 is merged

app/services/api/sectors/companies_grouped_by_level.rb
```diff
@@ -0,0 +1,88 @@
...
+          .map { |_level, companies| companies }
+          .flatten
+          .map { |company| series_options(company[:emissions], company[:name]) }
+      end
+
+      def series_options(data, series_name)
+        {
+          name: series_name,
+          data: data,
+          lineWidth: 4
```
kowal: Turns out that not all chart options can be passed in here when chart JSON data is downloaded **after** page load. So we either have to pre-render charts on the server or apply additional JS config.

app/services/api/charts/sector.rb
```diff
@@ -0,0 +1,129 @@
...
+      #       name: 'Below 2',
+      #       data: [ ['Coal Mining', 52], ['Steel', 73] ]
+      #     },
+      #     {
+      #       name: 'Paris',
+      #       data: [ ['Coal Mining', 65], ['Steel', 26] ]
+      #     }
+      #   ]
+      def group_by_cp_benchmark
+        CP_SCENARIOS.map do |cp_alignment|
```
kowal: FYI: this still uses random data

app/services/api/charts/company.rb
```diff
@@ -0,0 +1,88 @@
...
+    class Company
+      SECTORS_LEVELS_DESC = {
+        '0' => 'Unaware of Climate Change as a Business Issue',
+        '1' => 'Acknowledging Climate Change as a Business Issue',
+        '2' => 'Building Capacity',
+        '3' => 'Integrating into Operational Decision Making',
+        '4' => 'Strategic Assessment'
+      }.freeze
+
+      def details_data(company)
```
kowal: OK, so this isn't actually used by any chart. Maybe we could move it to regular views helpers.

app/services/api/charts/company.rb
```diff
@@ -0,0 +1,88 @@
...
+          name: cp_benchmark.scenario,
+          data: cp_benchmark.emissions
+        }
+      end
+
+      # returns average from sector companies
+      def emissions_data_series_from_sector(sector)
+        {
+          name: "#{sector.name} sector mean",
+          data: all_sector_assessments_emissions(sector).reduce(&:merge) # TODO: calculate real average from all sectors
```
kowal: working on this now...

app/helpers/chart_helper.rb
```diff
@@ -0,0 +1,34 @@
+module ChartHelper
```
agwozdowski: Can we move those methods to erb partial?

app/services/api/charts/company.rb
```diff
@@ -0,0 +1,110 @@
...
+    class Company
+      SECTORS_LEVELS_DESC = {
+        '0' => 'Unaware of Climate Change as a Business Issue',
+        '1' => 'Acknowledging Climate Change as a Business Issue',
+        '2' => 'Building Capacity',
+        '3' => 'Integrating into Operational Decision Making',
+        '4' => 'Strategic Assessment'
+      }.freeze
+
+      def details_data(company)
```
agwozdowski: Since you use `company` in almost all method, you can move `company` as an instance variable and assign it in initialize method

app/services/api/charts/company.rb
```diff
@@ -0,0 +1,110 @@
...
+
+        sector_average_for_year = lambda do |year|
+          company_emissions = all_sector_emissions.map { |emissions| emissions[year.to_s] }.compact
+          (company_emissions.sum / company_emissions.count).round(2)
+        end
+
+        years.map { |year| [year, sector_average_for_year.call(year)] }.to_h
+      end
+
+      def sector_all_emissions(sector)
```
agwozdowski: When you move `company` as an instance variable, you can memoize this method like:
```
def sector_all_emissions
    @sector_all_emissions ||=  @company.sector
        .companies
        .includes(:cp_assessments)
        .map(&:cp_assessments)
        .flatten.map(&:emissions)
end
```

kowal: fixed in fbefce8

app/services/api/charts/company.rb
kowal: fixed in fbefce8

app/controllers/sectors_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  # Section:  CP
+  # Type:     column chart
+  # On pages: :index
+  def benchmarks_chart_data
+    data = ::Api::Charts::Sector.new(companies_scope(params)).group_by_cp_benchmark
+
+    render json: data.chart_json
+  end
+
+  def companies_scope(params)
```
simaob: should `companies_scope` be inside a `private` block?

app/services/api/charts/company.rb
```diff
@@ -0,0 +1,121 @@
...
+        '4' => 'Strategic Assessment'
+      }.freeze
+
+      def initialize(company)
+        @company = company
+      end
+
+      def details_data
+        {
+          name: company.name,
```
simaob: missing the @ before `company`?

app/services/api/charts/company.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      private
+
+      def questions_by_level
+        company.latest_assessment.questions.group_by { |q| q['level'] }
+      end
+
+      def emissions_data_from_company
+        {
+          name: company.name,
```
simaob: I think we have to update to use `@company`, right?

config/routes.rb
```diff
@@ -1,4 +1,21 @@
 Rails.application.routes.draw do
+  # TPI
+  resources :sectors, only: [:show, :index] do
```
simaob: I think TPI and the future Laws website should be namespaced, as later they will sit in different domains, and might be easiest to link to /tpi or /laws, and then keep the routes isolated. Wondering if we should do that straight away or later when is actually needed. What do you think?

kowal: good point, I can add `/tpi` namespace now

app/services/api/charts/company.rb
kowal: I've added attr_accessor for this, but maybe using `@ivar` is better here.

Also, I'll move this `details_data` to another class (`Api::Presenter::Company` ?), as this is not related to charts at all.

simaob: it might be my rusty rails 😅 but because you setup `@company` on the initializer I thought you had to use that. Like you were doing in the sector.rb file with `@company_scope`

simaob: Yeah, maybe presenter would be a good place for that.

kowal: f7b5f29

app/controllers/sectors_controller.rb
kowal: d81af1b

config/routes.rb
kowal: d6efe5a

app/services/api/charts/sector.rb
kowal: @agwozdowski is working on that part (within this PR).

app/services/api/charts/company.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      def emissions_data_from_sector
+        {
+          name: "#{@company.sector.name} sector mean",
+          data: sector_average_emissions
+        }
+      end
+
+      def emissions_data_from_sector_benchmarks
+        @company.sector_benchmarks.map do |benchmark|
```
tsubik: for the chart sector benchmarks should be only for the last date less than CP::Assessment date. 
Ex. 
if we have benchmarks available for 04.2017 and 05.2018, and assessment date is 06.2018 we take 05.2018 benchmarks
if we have the same benchmarks but the assessment date is 06.2017 we take benchmarks for 04.2017

app/services/api/charts/company.rb
```diff
@@ -0,0 +1,94 @@
...
+      def sector_average_emission_for_year(year)
+        company_emissions = sector_all_emissions
+          .map { |emissions| emissions[year.to_s] }
+          .compact
+
+        (company_emissions.sum / company_emissions.count).round(2)
+      end
+
+      # Returns array of emissions ({ year => value }) from all Companies from current Sector
+      def sector_all_emissions
```
tsubik: We should only consider the last assessment for a given company

app/services/api/charts/sector.rb
tsubik: Ok, we cannot have CP_SCENARIOS as constant

config/routes.rb
```diff
@@ -1,4 +1,22 @@
 Rails.application.routes.draw do
+  namespace :tpi do
+    resources :sectors, only: [:show, :index] do
```
tsubik: should we have `API` namespace also? We can add that later too.

app/services/api/charts/sector.rb
```diff
@@ -0,0 +1,148 @@
...
+      #      [
+      #        { name: 'Alaska Air', status: 'up' },
+      #        { name: 'IAG', status: 'down' }
+      #      ]
+      #     ]
+      #   ]
+      def companies_summaries
+        @company_scope
+          .includes(:mq_assessments)
+          .group_by { |company| company.mq_assessments.order(:assessment_date).first.level }
```
tsubik: Company has `latest_assessment`, but I think I see a bug there, should use scope ordered by date.

tsubik: also should be renamed to `latest_mq_assessment`

tsubik: or `last_mq_assessment`

app/services/api/charts/sector.rb
```diff
@@ -0,0 +1,148 @@
...
+      def companies_count_per_sector_cp_scenarios(_cp_alignment)
+        ::Sector.pluck(:id, :name).map do |_sector_id, sector_name|
+          [sector_name, rand(100)]
+        end
+      end
+
+      def company_emissions_series_options(company)
+        {
+          name: company.name,
+          data: company.cp_assessments.last&.emissions,
```
tsubik: `last` could not be last here, as `has_many` scope is not ordered. We could add order by assessment_date directly to cp_assessments and mq_assessments relation's definitions. What do you think?

app/assets/stylesheets/tpi.scss
```diff
@@ -0,0 +1,19 @@
+.container {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  gap: 5px;
```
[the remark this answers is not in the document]
kowal: Turns out this is [FF only ](https://caniuse.com/#feat=flexbox-gap) but I think we can ignore it for now, @rapzo will probably rewrite all styling here anyway.

kowal:
@tsubik regarding `#group_by_cp_benchmark` and `CP_SCENARIOS` you're right, this is old code, from before our recent discussions. Adam is working on this now (grouping company numbers by sectors "scenario" alignments).

app/services/api/charts/sector.rb
kowal: yep, I think it's good idea to include this ordering in the scopes.

kowal: a306e70

app/services/api/charts/sector.rb
kowal: a306e70 (similar to your fix in https://github.com/Vizzuality/laws_and_pathways/pull/72)

config/routes.rb
kowal: Not sure what we would want to put in `api` namespace. Do you mean API for 2nd laws frontend?

app/services/api/charts/company.rb
kowal: fixed in 3745de5

app/services/api/charts/sector.rb
tsubik: :+1:

app/services/api/charts/sector.rb
tsubik: Right, I've tried to fix it in separate PR as you've seen and came into the problem of code clarity ;], like `latest_assessment` looks better when it is `mq_assessment.last` so the default order should be not descending. On the other side in the UI we want descending order.

tsubik: Weirdo, I thought that Rails will reset relation order when I want to override it, like when I have order on relation and then want to reverse it like using `latest_first` that does not work. So, I think we should not add order on relation and keep the code as it is here https://github.com/Vizzuality/laws_and_pathways/pull/72

kowal: ok, I moved the ordering from scope to method and added test case for it - 055dfa4.

app/services/api/charts/company.rb
kowal: ok, so few thoughts on this:

1) the current implementation is clearly wrong because we can have multiple CP benchmarks per scenario and we don't want to see two or three `Paris Pledges` series on a single chart.

2) **CP Company chart** on mockups has `Assessment Date` select input which provides scope for company emissions series ... but does it necessarily mean that we shouldn't always show the latest CP Benchmarks? I kind of understand what's @tsubik point here (that sometimes it may be useful to show benchmarks from before last assessment) but want to confirm that.

3) the previous point is only about the **CP Company chart**. For **CP Sector chart** we don't have a single assessment context so we will always show most recent CP benchmarks, right?

/cc @simaob @agwozdowski @rapzo

kowal: I've implemented what @tsubik suggested in https://github.com/Vizzuality/laws_and_pathways/pull/75 for **CP Company chart** (also fixed CP Sector chart there).

kowal: fixed in #75

PR #58, opened 2019-09-16
Seed sectors (names and CP benchmark units)

kowal:
Units are taken from http://www.lse.ac.uk/GranthamInstitute/tpi/the-toolkit/

Zrzut ekranu 2019-09-17 o 08 32 11
Zrzut ekranu 2019-09-17 o 08 32 24

### DIFF
--- a/app/admin/sectors.rb
+++ b/app/admin/sectors.rb
@@ -11,6 +11,7 @@ ActiveAdmin.register Sector do
     column :name do |sector|
       link_to sector.name, admin_sector_path(sector)
     end
+    column 'CP Benchmark unit', &:cp_unit
     column :slug
     actions
   end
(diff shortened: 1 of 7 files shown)


db/seeds.rb
```diff
@@ -1 +1,26 @@
-AdminUser.create!(email: 'admin@example.com', password: 'password', password_confirmation: 'password') if Rails.env.development?
+if Rails.env.development?
```
simaob: Should we move the Units to outside the development env? As we'll need them in prod and staging as well.

app/admin/sectors.rb
```diff
@@ -58,6 +60,7 @@
 
     f.inputs do
       f.input :name
+      f.input :cp_unit, label: 'Carbon Performance benchmark unit'
```
tsubik: should we just call it `Carbon Performance Unit` ?

kowal: year, works for me :) @simaob ?

db/seeds.rb
kowal: Right! Also, since we can edit them in the interface I'll add a condition to only set units from seeds for new records.

PR #59, opened 2019-09-17
Feature/import sector benchmarks

tsubik:
1. Changing the structure of CP::Benchmarks

Previously CP::Benchmark has `benchmarks` jsonb column with different scenarios and emissions values for them. I decided to flatten that structure to incorportate `scenario` in the model and `emissions` jsonb. Import/export and data management for single benchmark is easier now, it also makes more sense.

2. Import/export of CP Benchmarks on the new admin page for benchmark management.

image

3. Edit form for CP Benchmark with dynamic emissions table

image

To reimport existing benchmarks

```
bundle exec rake import:cp_benchmarks
```


kowal:
When adding new emission values like this:
Zrzut ekranu 2019-09-17 o 12 20 50
.. no error is raised and the value is not added. I think we should either show an error or allow to update it without the emission.

tsubik:
Hmm, I think it worked without providing emission, let me check

tsubik:
Ok, @kowal I fixed that.

app/admin/cp_benchmarks.rb
```diff
@@ -0,0 +1,58 @@
...
+      row :updated_at
+    end
+
+    panel 'Benchmark emission values' do
+      render 'admin/cp/emissions_table', emissions: resource.emissions
+    end
+  end
+
+  csv do
+    year_columns = collection.flat_map(&:emissions_all_years).uniq.sort
```
kowal: It might be more readable to extract this somewhere. Maybe some decorator / helper?

app/admin/sectors.rb
```diff
@@ -35,15 +35,17 @@
               'No Carbon Performance Benchmarks for this sector yet'
             end
           else
-            resource.cp_benchmarks.latest_first.map do |benchmark|
-              panel "Released in #{benchmark.date.strftime('%B %Y')}", class: 'benchmark' do
-                table_for benchmark.benchmarks, class: 'cell-padding-sm cell-centered' do
-                  column :name do |b|
-                    b['name']
+            resource.cp_benchmarks.latest_first.group_by(&:release_date).map do |release_date, benchmarks|
```
kowal: here as well, some logic could be extracted to view helpers

app/models/cp/benchmark.rb
```diff
@@ -2,25 +2,41 @@
...
+    validates_presence_of :release_date, :scenario
 
-    def benchmarks_all_years
-      benchmarks.map { |b| b['values'].keys }.flatten.uniq
+    def emissions_all_years
+      emissions.keys
+    end
+
+    def average_emission
+      not_empty_values = emissions.values.compact
```
kowal: `none_empty_values`?

app/services/csv_import/cp_benchmarks.rb
```diff
@@ -0,0 +1,65 @@
...
+    def emissions(row)
+      row.headers.grep(/\d{4}/).map do |year|
+        {year.to_s.to_i => parse_float(row[year])}
+      end.reduce(&:merge)
+    end
+
+    def parse_float(value)
+      return unless value.present?
+
+      value.to_f
```
kowal: `value&.to_f` but maybe current code is more readable..

app/services/csv_import/cp_benchmarks.rb
```diff
@@ -0,0 +1,65 @@
...
+      }
+    end
+
+    def parse_date(date)
+      Import::DateUtils.safe_parse(date, ['%Y-%m', '%Y-%m-%d'])
+    end
+
+    def emissions(row)
+      row.headers.grep(/\d{4}/).map do |year|
+        {year.to_s.to_i => parse_float(row[year])}
```
kowal:
```ruby
row.headers.grep(EMISSION_VALUE_PATTERN)
  .map { |year| {year.to_i => parse_float(row[year])} } # to_s is really needed?
  .reduce(&:merge)
```
Also I got a feeling we can replace `.map {}.reduce { }` with just one `.reduce { }`.

app/models/cp/benchmark.rb
```diff
@@ -2,25 +2,41 @@
...
   class Benchmark < ApplicationRecord
     belongs_to :sector
 
-    scope :latest_first, -> { order(date: :desc) }
-    scope :by_date, -> { order(:date) }
+    scope :latest_first, -> { order(release_date: :desc) }
+    scope :by_release_date, -> { order(:release_date) }
 
-    validates_presence_of :date
+    validates_presence_of :release_date, :scenario
```
kowal: Maybe we should validate `scenario` with inclusion on the known scenario list (https://github.com/Vizzuality/laws_and_pathways/pull/57/files#diff-bcd1663b967ed112eee68c615f3ac627R4)?

@simaob do you know if that list (of scenarios) should be open or rather strict? Do we need to manage this list as well?

app/services/csv_import/cp_benchmarks.rb
tsubik: I will change it back, actually, it does not make sense to keep empty values. The same I will change in the interface, empty values will be discarded when saving.

app/services/csv_import/cp_benchmarks.rb
tsubik: true, good point

app/models/cp/benchmark.rb
tsubik: I would keep that as just a name with the flexibility to change to whatever. They already have different names for different release dates and sectors. As long we don't need any complex reports it should be fine.

kowal: hmm but later we will need only a known list of scenario on the frontend. How will we map it if there are more scenario types in the DB?

Zrzut ekranu 2019-09-18 o 12 17 42

tsubik: You only need names, and those benchmarks need to be grouped by `release_date` and only one benchmark group should be taken (everything depends on assessment date - an appropriate benchmark should be taken by date). Scenario names will be from that chosen benchmark group.

simaob: Not sure if the colour has any meaning, I can try to double check with Fausto. If not we can have a number of colours that are flexible enough to an expected number of scenarios and then sort them always the same way, so that they always get the same colour. Something like that?

tsubik: We can show data against only one benchmark set/group.

tsubik: If the color has meaning from the lowest to the highest scenario we can add colors based on the average of emission values, scenario name won't matter that much.

simaob: oh right, good point!

app/admin/cp_benchmarks.rb
tsubik: Hmm, just this one line? not sure if that would help for me this is readable

app/services/csv_import/cp_benchmarks.rb
tsubik: not sure why `to_s` is here, maybe it was needed. I copied that from climate watch, but I can remove it, we will see :P

app/models/cp/benchmark.rb
tsubik: @kowal is right here, on the image below I can only see the top of the chart, but looks like that chart needs standardized scenario names across different sectors. They are not standardized now, so I believe in the next iteration when we know that all scenarios are the same we can add validation to be only those from strict list.

app/services/csv_import/cp_benchmarks.rb
tsubik: ok, I know why `to_s` year is a symbol.

PR #62, opened 2019-09-18
Feature/litigation soft delete

agwozdowski:
1. Added discarded_at to `Litigation` model
2. Override `DESTROY` action using `::Command::Destroy::Litigation`
3. Added specs
4. Removed validation during soft-delete action in Litigation and also from Legislation because it's possible that some precedent record (like geography) was removed earlier.

### DIFF
--- a/app/commands/command/destroy/legislation.rb
+++ b/app/commands/command/destroy/legislation.rb
@@ -15,10 +15,6 @@ module Command
 
             r.litigations = []
             r.targets = []
-
-            # Validation is false because it's possible that some precedent
-            # record (like geography) was removed earlier
-            r.save(validate: false)
           end
         end
       end
--- a/app/commands/command/destroy/legislation.rb
+++ b/app/commands/command/destroy/legislation.rb
@@ -15,7 +15,10 @@ module Command
 
             r.litigations = []
             r.targets = []
-            r.save!
+
+            # Validation is false because it's possible that some precedent
+            # record (like geography) was removed earlier
+            r.save(validate: false)
           end
         end
       end
--- a/app/admin/geographies.rb
+++ b/app/admin/geographies.rb
@@ -88,7 +88,7 @@ ActiveAdmin.register Geography do
                   {alert: 'Could not delete selected Geography'}
                 end
 
-      redirect_to admin_geographies_path(scope: current_scope.name), results
+      redirect_to admin_geographies_path, results
     end
   end
 end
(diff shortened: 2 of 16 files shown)


app/commands/command/destroy/legislation.rb
```diff
@@ -15,7 +15,10 @@ def call
 
             r.litigations = []
             r.targets = []
-            r.save!
+
+            # Validation is false because it's possible that some precedent
+            # record (like geography) was removed earlier
+            r.save(validate: false)
```
kowal: to be honest, do we need this save here at all? I know I've added it in the first PR from this series, but need to re-check it 😅

agwozdowski: I've just checked it, and it works without save :)

kowal: nice!

tsubik:
I'm on it. :eyes:

agwozdowski:
Thanks @tsubik ! I'll take a look and fix in all places

agwozdowski:
@kowal @tsubik 
I made a research with this redirect. It doesn't work on all soft-delete pages.
Solution with `current_scope` which was used at the beginning will not work because `current_scope` is always `all` in this case.

I tried `redirect_to request.referer` which works properly when user delete it from the list, but it doesn't work from view page when user click DELETE. 
 
We have 2 options:
1. The easiest - always redirect to `all` scope
2. Redirect to last used scope. This solution is more complicated because by default DELETE in active admin doesn't send last scope. It  calls `http://localhost:3000/admin/geographies/21` with param `_method: delete`. It's probably doable, but I'm not sure if it's necessary. Before I'll start investigation how override this action want to be sure if want to invest time for it.

tsubik:
Hmm, I think default ActiveAdmin destroy must not have this issue. Maybe it's worth to check the AA code how they redirecting after destroy.

agwozdowski:
Hey @tsubik, @kowal  - it's again me :)
I've just checked that in "normal" Destroy - without our overwrite - active admin redirect to the list without any scope. 
My suggestion is to implement it similar in our case in all lists.
What do you think?

kowal:
Fine for me, let's not block this story (since the problem was there before). We can improve those redirect later. @agwozdowski can you add a pivotal task to track that?

agwozdowski:
@kowal added a story but it's a minor in my opinion. For now it's working in the same way like active-admin DESTROY action

PR #64, opened 2019-09-19
Feature/company soft delete

agwozdowski:
Soft delete for Company model. Added also `discarded_at` to `mq_assessments` and `cp_assessments`

### DIFF
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -176,5 +176,17 @@ ActiveAdmin.register Company do
     def scoped_collection
       super.includes(:geography, :headquarters_geography, :mq_assessments)
     end
+
+    def destroy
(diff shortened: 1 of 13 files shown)


app/admin/companies.rb
```diff
@@ -176,5 +176,17 @@
...
+    def destroy
+      destroy_command = ::Command::Destroy::Company.new(resource.object)
+
+      results = if destroy_command.call
+                  {notice: 'Successfully deleted selected Company'}
+                else
+                  {alert: 'Could not delete selected Company'}
+                end
+
+      redirect_to admin_companies_path, results
```
kowal: What do you think about changing `results` to `status` or `message` in all cases like this?

agwozdowski: 👍 changed to `message` for delete, archive and batch actions

PR #65, opened 2019-09-20
Feature/target soft delete

agwozdowski:


app/admin/target_scopes.rb
```diff
@@ -1,7 +1,46 @@
...
+    f.inputs do
+      f.input :name
+    end
+
+    f.actions
+  end
+
+  controller do
+    def destroy
+      destroy_command = ::Command::Destroy::TargetScope.new(resource)
```
kowal: Maybe it's worth to extract this whole action to some module (parametrized with resource class). It looks like it won't change much anymore.

agwozdowski: @kowal added DiscardableController. Also moved DiscardableModel to concerns. Take a look please

app/controllers/concerns/discardable_controller.rb
```diff
@@ -0,0 +1,16 @@
+module DiscardableController
+  # Override default Active Admin destroy method using Discardable Model.
+  # It sets discarded_at for entity and for its dependent resources.
+  # Redirect to resource list with proper message
+  def destroy
+    entity_name = ActiveSupport::Inflector.singularize(controller_name).camelize
+    entity_destroy_class = "::Command::Destroy::#{entity_name}".constantize
```
kowal: can we catch possible `NameError`s here and re-throw ex with more context?

agwozdowski: @kowal done

PR #67, opened 2019-09-24
First version of "Carbon Performance: All sectors" chart with real data

agwozdowski:
- First version of `Carbon Performance: All sectors` chart with real data. 
- Unit tests for `Carbon Performance: All sectors` chart service
- Unit tests for Happy Path all chart services
- Small refactoring for some services


simaob:
On which path should I be able to see this? And do I need to run any import tasks? I'm getting an internal server error when trying to load the chart on this page: http://localhost:3000/tpi/sectors

agwozdowski:
@simaob I run it locally. Before it, I run db setup and importer with
`rake db:setup` and `rake import:all`.
Chart looks like:
chart
I'll check data again tomorrow because I'm not sure if they are correct.

app/services/api/charts/cp_benchmark.rb
```diff
@@ -0,0 +1,117 @@
...
+          sector.companies.each do |company|
+            next if scenarios.empty?
+
+            company_emission = company_emission(company)
+
+            next unless company_emission
+
+            company_scenario = company_scenario(company_emission, scenarios)
+
+            results[company_scenario] ||= Hash.new(0)
```
kowal: My main concern here is that because we start building this data from **sectors**, we have to perform this hash-based aggregation/counting in 2 places.

Please check https://github.com/Vizzuality/laws_and_pathways/pull/69 - although it still doesn't work, maybe we can give it a try?:)

agwozdowski: @kowal it start building from `sectors`. `company_scenario` here means which sector scenario match to current emission for company.
I'll take a look on your pull request :)

app/services/api/charts/cp_benchmark.rb
```diff
@@ -0,0 +1,117 @@
...
+      # Get emissions for all scenarios for all sectors
+      # @return [Hash]
+      # @example
+      #   {"Airlines": {"Below 2 Degrees": 110, "2 Degrees": 101}}
+      def sectors_scenario_emissions
+        sectors = {}
+
+        ::Sector.all.each do |sector|
+          sectors[sector] = {}
+          sector.cp_benchmarks.each do |benchmark|
```
tsubik: Hmm, only the latest benchmarks should be considered here

agwozdowski: Fixed here:
https://github.com/Vizzuality/laws_and_pathways/pull/67/commits/94ebc72ebb803484045c902751399f2fdd1ccc3b

app/services/api/charts/cp_benchmark.rb
```diff
@@ -0,0 +1,128 @@
...
+      end
+
+      # Get emissions for all scenarios for all sectors
+      # @return [Hash]
+      # @example
+      #   {"Airlines": {"Below 2 Degrees": 110, "2 Degrees": 101}}
+      def sectors_scenario_emissions
+        sectors = {}
+
+        ::Sector.all.each do |sector|
```
kowal: I think we should include (eager load) all required relation here (assessments, benchmarks).

app/services/api/charts/cp_benchmark.rb
```diff
@@ -0,0 +1,128 @@
...
+      #       "International Pledges" => {
+      #       "Steel" => 80
+      #     }
+      #   }
+      def sectors_cp_performance
+        results = {}
+
+        sectors_scenario_emissions.each do |sector, scenarios|
+          sector.companies.each do |company|
+            next if scenarios.empty?
```
kowal: this `next` should be in the outer loop, right?

kowal: or maybe we can remove it if `sectors_scenario_emissions` won't return empty scenarios?

app/services/api/charts/cp_benchmark.rb
```diff
@@ -0,0 +1,128 @@
...
+        end
+
+        sectors
+      end
+
+      # Get company emission for current year or for the latest assessment
+      # @param company [Company]
+      # @return [Float]
+      def company_emission(company)
+        emissions = company.cp_assessments.order(:assessment_date).last&.emissions
```
kowal: we can now use `company.latest_cp_assessment.emissions`

app/services/api/charts/cp_benchmark.rb
agwozdowski: I see that it's not necessary. I'll remove this condition

PR #69, opened 2019-09-26
(chart-api) less db queries

kowal:
Since this one is quite complicated, instead of a regular review I decided to also give it a shot and take a different approach to prepare data for this chart (and to understand the problem better).

@agwozdowski Some tests are still failing :/ probably I broke smth related to picking correct emission date.


kowal:
Closing this one. I will try to get back to this after we do more manual tests of the frontend and have more specs in place.

PR #71, opened 2019-09-26
adding country flags

tsubik:
I'm not sure if we should also or only allow flag image import or just providing flags is good enough.

image


tsubik:
Going to incorporate script to update flags here.

tsubik:
Rake task `flags:generate` to regenerate country flags has been added.

PR #72, opened 2019-09-26
fixing mq assessment order, adding more colors

tsubik:
`latest_assessment` for a company was not always the latest one, because order by date was missing somehow. That affected current company level, but now should be fixed. Also I added more colors to the backend (not sure if my idea is good lol)

image

### DIFF
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -25,7 +25,7 @@ ActiveAdmin.register Company do
   sidebar 'Details', only: :show do
     attributes_table do
       row :company, &:name
-      row :level, &:mq_status_description_short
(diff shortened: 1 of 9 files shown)


kowal:
that might be not related, but one test failed:
```
  1) Admin::TargetsController POST create with valid params creates a new Target
     Failure/Error: expect(t.legislation_ids).to eq(legislations.pluck(:id))
     
       expected: [105, 106]
            got: [106, 105]
     
       (compared using ==)
     # ./spec/controllers/admin/targets_controller_spec.rb:109:in `block (5 levels) in <top (required)>'
```

tsubik:
I will fix that. Thanks :+1:

PR #74, opened 2019-09-28
Feature/instrument type and instrument

agwozdowski:
- Added InstrumentType to "Laws" menu. Implemented also soft-delete for it.
- Instrument model, migration, resource, form, soft-delete
- Added instruments to Legislation model and views

### DIFF
--- a/app/decorators/legislation_decorator.rb
+++ b/app/decorators/legislation_decorator.rb
@@ -43,8 +43,6 @@ class LegislationDecorator < Draper::Decorator
   end
 
   def instrument_links
-    return [] if model.instruments.empty?
-
     model.instruments.map do |instrument|
       h.link_to instrument.name,
                 h.admin_instrument_path(instrument),
--- a/app/decorators/legislation_decorator.rb
+++ b/app/decorators/legislation_decorator.rb
@@ -41,4 +41,15 @@ class LegislationDecorator < Draper::Decorator
                 title: document.external? ? document.url : nil
     end
   end
+
+  def instrument_links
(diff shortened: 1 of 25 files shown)


app/decorators/legislation_decorator.rb
```diff
@@ -41,4 +41,15 @@ def document_links
                 title: document.external? ? document.url : nil
     end
   end
+
+  def instrument_links
+    return [] if model.instruments.empty?
```
kowal: I think you want to check `model.instruments.nil?` here or remove this check completely because `instruments.map` will handle empty array.

agwozdowski: Removed this if. Take a look please

spec/controllers/admin/instrument_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'rails_helper'
+
+RSpec.describe Admin::InstrumentsController, type: :controller do
+  let(:admin) { create(:admin_user) }
+  before { sign_in admin }
+
+  describe 'DELETE destroy' do
+    let!(:instrument) { create(:instrument, discarded_at: nil) }
+
+    context 'with valid pspec/factories/instrument_types.rb:16arams' do
```
tsubik: some accidental ctrl-v here

kowal: :male_detective:

agwozdowski: :D thanks @tsubik !

PR #75, opened 2019-09-29
(chart-api) fix for fetching sector benchmarks + clean up

kowal:
Extracted query helpers:
- `Sector#latest_released_benchmarks`
- `Company#latest_sector_benchmarks`
- `Company#latest_sector_benchmarks_before_last_assessment`

Also, updated docs.

### DIFF
--- a/app/models/company.rb
+++ b/app/models/company.rb
@@ -59,21 +59,29 @@ class Company < ApplicationRecord
     sector.latest_released_benchmarks
   end
 
-  # Returns sector benchmarks:
-  # - if company has assessment - from the last date before latest CP::Assessment date
-  # - if company has no assessments - from latest benchmarks
+  # Returns sector CP benchmarks:
+  # - if company has assessment      - from the last date before latest CP::Assessment date
+  # - if company has no assessments  - from latest benchmarks
+  # - if company has old assessments - from latest benchmarks
+  #
+  # @return [AssociationRelation [#<CP::Benchmark]
   #
   # @example Company has assessment:
   # - benchmarks available for 04.2017 and 05.2018
-  # - if assessment date is 06.2018 we take 05.2018 benchmarks
-  # - if assessment date is 06.2017 we take benchmarks for 04.2017
+  # - if assessment date is 06.2018 - we take benchmarks from 05.2018
+  # - if assessment date is 06.2017 - we take benchmarks from 04.2017
   def latest_sector_benchmarks_before_last_assessment
     last_assessment_date = latest_cp_assessment&.assessment_date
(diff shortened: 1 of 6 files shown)


app/models/company.rb
```diff
@@ -55,7 +55,37 @@ def latest_cp_assessment
...
+  #
+  # @example Company has assessment:
+  # - benchmarks available for 04.2017 and 05.2018
+  # - if assessment date is 06.2018 - we take benchmarks from 05.2018
+  # - if assessment date is 06.2017 - we take benchmarks from 04.2017
+  def latest_sector_benchmarks_before_last_assessment
+    last_assessment_date = latest_cp_assessment&.assessment_date
+    sector_benchmarks_dates = sector.cp_benchmarks.pluck(:release_date).uniq.sort
+
+    if last_assessment_date
```
agwozdowski: I prefer to use early return instead big if, like:
```
unless last_assessment_date
  return 

last_release_date_before_assessment =
....
```

app/models/sector.rb
```diff
@@ -19,4 +19,8 @@ class Sector < ApplicationRecord
 
   validates_presence_of :name, :slug
   validates_uniqueness_of :name, :slug
+
+  def latest_released_benchmarks
```
tsubik: :+1:

spec/models/company_spec.rb
```diff
@@ -86,4 +86,63 @@
       expect(company.latest_mq_assessment.assessment_date.to_s).to eq('2019-05-01')
     end
   end
+
+  describe '#latest_sector_benchmarks_before_last_assessment' do
```
tsubik: :+1:

app/models/company.rb
```diff
@@ -55,7 +55,36 @@ def latest_cp_assessment
...
-  def sector_benchmarks
-    sector.cp_benchmarks
+  def latest_sector_benchmarks
+    sector.latest_released_benchmarks
+  end
+
+  # Returns sector CP benchmarks:
+  # - if company has assessment      - from the last date before latest CP::Assessment date
+  # - if company has no assessments  - from latest benchmarks
+  # - if company has old assessments - from latest benchmarks
```
tsubik: what does it mean "old assessments" here?

kowal: old = "from before the oldest benchmark", I'll update it.

tsubik: so then the oldest benchmark should be used, right?

tsubik: ahh, no, then I don't know what should be used :)

kowal:
waiting from @simaob confirmation if that's what we want here :)

simaob:
The updated comment looks good to me and makes sense! Thanks for sorting it out.

PR #78, opened 2019-10-07
add last reported year to cp assessments

tsubik:
Small PR to add last reported year to CP::Assessment model. This year is a boundary between reported and targeted emission data.


tsubik:
Done. Thanks for reminding me.

PR #80, opened 2019-10-14
Feature/ 'assessments levels over the years' chart

kowal:
### Summary
Add Assessment Levels/years chart for (TPI) Management Quality for Company 

### Details

* New endpoint returns data in format: `[[2016,2],[2017,2],[2018,4]]`
* `::Api::Charts::Company.new(@company).nr_of_assessments_data` always provides data for each year starting from latest year to most recent, filling missing years levels with last reported level.

### Mockups
tpi-mocks

### Example
More than 1 assessment (http://localhost:3000/tpi/companies/314)

tpi-levels

When only 1 assessment (http://localhost:3000/tpi/companies/460)

tpi-2

### Questions

* should the chart X axis end on last reported year or current year? **Yes**
* what to do with companies which has only single MQ assessment in current year? We can't draw line in that case, so chart doesn't make sense. **Draw a dot**


simaob:
Thanks for this Kowal! =) I think that we should show current year as the latest year on the xx axis, this way it gives more context and can make it easier for people to see if companies submitted something this year.

One there's only one MQ Assessment we can just show a single dot, no? I think it's fair.

kowal:
> Thanks for this Kowal! =) I think that we should show current year as the latest year on the xx axis, this way it gives more context and can make it easier for people to see if companies submitted something this year.

Yep, make sense, so I need to fix that in this PR

> 
> One there's only one MQ Assessment we can just show a single dot, no? I think it's fair.

OK, that's how it works now.

kowal:
> > .. I think that we should show current year as the latest year on the xx axis, this way it gives more context and can make it easier for people to see if companies submitted something this year.
> 
> Yep, make sense, so I need to fix that in this PR

.. although it might make sense to somehow distinguish is vizzualy. This case means that we don't have data for those last year(s), but we're assuming MQ a. level didn't change - but it could - company could have done better/worst job but we just don't know it.

kowal:
@simaob I fixed the chart to always end on the current year, also I simplified the logic. I still need to add tests for this.

app/controllers/tpi/companies_controller.rb
```diff
@@ -3,7 +3,20 @@ class CompaniesController < ApplicationController
...
+      @company_summary = company_presenter.summary
+      @company_mq_assessments = company_presenter.mq_assessments
+    end
+
+    # Data:     Company MQ Assessments Levels over the years
+    # Section:  MQ
+    # Type:     line chart
+    # On pages: :show
+    def assessments_levels_chart_data
+      @company = Company.find(params[:id])
```
tsubik: Maybe we could move that into `before_action :fetch_company`

app/services/api/presenters/company.rb
```diff
@@ -13,15 +13,20 @@ def initialize(company)
         @company = company
       end
 
-      def company_details
+      def summary
         {
           name: @company.name,
           country: @company.geography.name,
           sector: @company.sector.name,
           market_cap: 'Large',
```
tsubik: This is just company.size

kowal:
> I wonder if we should take a year from an assessment date or publication date. Also what happens when there are two assessments/publications in the given year. Take a look at company `Bukit Asam`, the level is 1 because in the same year there were two assessments. But the second assessment was published the next year.
> 
> Maybe we should take publication date for this chart.

Great catch! will fix that in the next days. I think switching to publication date make sense. @simaob ?

tsubik:
Or we should just go more granular there. Not years, but month-year for axis-x

tsubik:
There could be 2 publications in one year also.

app/controllers/tpi/companies_controller.rb
kowal: :railway_track: fixed

app/services/api/presenters/company.rb
kowal: updated

tsubik:
I only show the points for real assessments. I removed that assumption that level stays the same when there was no assessment in the given year.

image

So the chart looks like this now. The number of data points is the number of assessments.

Do you think that is a good approach @simaob ?

tsubik:
Also, I'm using slugs to access companies and sectors on the public site. http://localhost:3000/tpi/companies/acerinox, http://localhost:3000/tpi/companies/bukit-asam

PR #81, opened 2019-10-15
Feature/missing uploaders

tsubik:
I was planning to get rid of AWS in this PR, but I will do that tomorrow in a separate one.

Couple things here:
- data uploader for geographies and changing the default CSV export template
- missing keywords and natural hazards import/export for legislations
- Events uploader
- Events index page and also show/edit/delete
- Add visibility status and legislation ids to Litigation importer

### DIFF
--- a/spec/commands/csv_data_upload_spec.rb
+++ b/spec/commands/csv_data_upload_spec.rb
@@ -4,7 +4,6 @@ require 'rails_helper'
 
 describe 'CsvDataUpload (integration)' do
   let(:legislations_csv) { fixture_file('legislations.csv') }
-  let(:litigations_csv) { fixture_file('litigations.csv') }
   let(:companies_csv) { fixture_file('companies.csv') }
   let(:targets_csv) { fixture_file('targets.csv') }
   let(:cp_benchmarks_csv) { fixture_file('cp_benchmarks.csv') }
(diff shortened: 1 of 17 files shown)


spec/commands/csv_data_upload_spec.rb
```diff
@@ -49,14 +49,53 @@
       legislations_csv,
       new_records: 2, not_changed_records: 0, rows: 2, updated_records: 0
     )
+
+    legislation = Legislation.find_by(title: 'Climate Law')
+
+    expect(legislation.legislation_type).to eq('legislative')
```
kowal: maybe `expect(legislation).to have_attributes(..)`?

spec/commands/csv_data_upload_spec.rb
```diff
@@ -49,14 +49,53 @@
...
+    expect(legislation.natural_hazards.size).to eq(2)
+    expect(legislation.natural_hazards_list).to include('sharkinados', 'flooding')
   end
 
   it 'imports CSV files with Litigation data' do
+    legislation1 = create(:legislation)
+    legislation2 = create(:legislation)
+    updated_litigation = create(:litigation)
+
+    csv_content = <<-CSV
```
kowal: I understand that this needs to be dynamic here, but can we still keep this in `let`-vars?

spec/commands/csv_data_upload_spec.rb
```diff
@@ -49,14 +49,53 @@
...
     expect_data_upload_results(
       Litigation,
       litigations_csv,
-      new_records: 4, not_changed_records: 0, rows: 4, updated_records: 0
+      new_records: 1, not_changed_records: 0, rows: 2, updated_records: 1
     )
+
+    litigation = Litigation.find_by(title: 'Litigation number 1')
+
+    expect(litigation.jurisdiction.iso).to eq('GBR')
```
kowal: same here: `have_attributes()` matcher could be more readable.

spec/commands/csv_data_upload_spec.rb
```diff
@@ -194,19 +268,54 @@
...
       expect(command.details.symbolize_keys).to eq(expected_details)
     end.to change(uploaded_resource_klass, :count).by(expected_details[:new_records])
   end
 
-  def fixture_file(filename)
+  def fixture_file(filename, content: nil)
+    file_path = "#{Rails.root}/spec/support/fixtures/files/#{filename}"
+
+    if content.present?
+      file_path = "#{Rails.root}/tmp/litigation_sides.csv"
```
kowal: hmm, why we have reference to `litigation_sides` in generic `fixture_file` helper?:)

spec/commands/csv_data_upload_spec.rb
tsubik: hmm, have_attributes nice idea :+1:

spec/commands/csv_data_upload_spec.rb
tsubik: I will see if that will be more readable, maybe if we split those tests to separate files to test only importer services.

spec/commands/csv_data_upload_spec.rb
tsubik: oh shit, good catch. That file name for tmp file does not actually matter but should be `tmp/#{filename}`

spec/commands/csv_data_upload_spec.rb
kowal: yeah, maybe we can move to `let`s just things like `legislation2` , `updated_litigation`?

tsubik: Hmm, moving those values to `let` globally would make test less readable in this case. Everything declared directly in test makes things easier when the test file is too big. There should be either `context` or separate file, but I don't think we have time to refactor tests much.

tsubik: Lol, I should have picked MiniTest instead of RSpec, fewer things to remember. Minimalism FTW!

PR #84, opened 2019-10-23
Feature/sectors clean up

tsubik:
- Changing `Sector` -> `TPISector`
- Adding new `LawsSector` for Targets, Litigation, Legislation
- Adding Laws sector admin page
- Changing uploaders for Litigation, Legislation and Targets to use `LawsSector`


simaob:
I might be doing something wrong but I'm getting a `NameError: uninitialized constant RenameSectorsToTPISectors` when running `rails db:migrate`

tsubik:
@simaob fixed. I ran that migration and after that added inflection for `TPI`.

PR #87, opened 2019-10-27
Feature/geographies changes

tsubik:
Couple changes for Geography model

- New geography types, adding geography type selection to the UI (https://www.pivotaltracker.com/story/show/169121339)
- Removing INDC url column and generating the UNFCCC link (https://www.pivotaltracker.com/story/show/169121318)
- Remove extraneous geography relation from Litigations (https://www.pivotaltracker.com/story/show/169190498)
- Rename Region -> World Bank Region (only UI and CSV file) (https://www.pivotaltracker.com/story/show/169121365)
- ISO code is mandatory only for National type (https://www.pivotaltracker.com/story/show/169121330)


tsubik:
@simaob wait going to remove `approach_to_climate_change` as per latest notes that will be a publication

tsubik:
Actually, it was only as a column in the table, maybe it was planned to save it but not display it. That only brings confusion, as they will do that as a publication, removing that field is the best option.

PR #90, opened 2019-11-06
Add react rails

aabdaab:
This small PR sets up `react-rails` gem to use React for more interactive widgets like `Bubble Chart`.
Read [README](https://github.com/reactjs/react-rails) for more info but long story short, that allows us to create react components in `app/javascript/components` and use them in `erb` files like so:
```
<%= react_component("HelloWorld", { greeting: "Hello from react-rails." }) %>
```
The first argument being component's name and second being `props`. HelloWorld added for demonstration.

I also added the number of companies in the center of the PieChart (not ideally styled yet) `/tpi/sectors` - check here


rapzo:
Must we do it The Rails Way™? We could achieve the same with a PWA imho.
Personal preference, is all.

simaob:
given the time pressure I'd go with something we know already and are comfortable with. Reduce the risks as much as possible.

aabdaab:
The Rails Way was precisely what I wanted to avoid 😅 That's why I added `react-rails`, what do you mean exactly?

tsubik:
Ok, let's do BubbleChart component with some default text for now instead of HelloWorld one.

aabdaab:
@tsubik done 👍

aabdaab:
done

PR #91, opened 2019-11-06
Feature/dynamic tpi companies assessment change

tsubik:
Ready for Review. We can change both CP and MQ assessment dates for a company. Please write if something does not work.


tsubik:
Lol, sometimes I see wrong charts are loaded 
image

Not sure if this is some problem with the marriage of turbolinks and chartkick, but I will try to fix it later. For now, I will disable turbolinks.

simaob:
I'm looking at a company with two assessments and when I select the oldest one the chart only shows that one. Shouldn't it still show two? Maybe we could have an "active" state on the dot of the chart, to indicate which one is selected. Or a different colour for that dot, is this possible?

tsubik:
Which company, that should work :]

tsubik:
Got tests for that!

simaob:
http://localhost:3000/tpi/companies/african-rainbow-minerals

simaob:
two dots if I selected the 2018:

Screenshot 2019-11-06 at 19 47 55

One dot if I select the 2017:
Screenshot 2019-11-06 at 19 47 49

tsubik:
Ahh, yes. That works, only showing up to that point.

tsubik:
I will try to figure out how to create some active point. Maybe a separate story for that?

simaob:
But shouldn't it show all the points, always?
And yes on the "active" we can create a new story for that.

tsubik:
I think you are right, here it makes sense. I thought it's gonna show how it was before, as a historic view.

tsubik:
Will do it tomorrow morning.

app/javascript/packs/tpi.js
```diff
@@ -0,0 +1,14 @@
+import Turbolinks from "turbolinks";
+import Chartkick from "chartkick";
+import HighCharts from "highcharts";
+import Rails from "rails-ujs";
```
rapzo: We should review this later since this stuff adds jQuery to the bundle and, with React already in, it's useless and waaay to much to have in the bundle.

app/javascript/tpi/controllers/index.js
```diff
@@ -0,0 +1,9 @@
+// Load all the controllers within this directory and all subdirectories.
+// Controller files must be named *_controller.js.
+
+import { Application } from "stimulus";
```
rapzo: Maybe we can drop this stuff as well, yeah?

tsubik: Right, I think there will be controllers eventually, so this one can stay.

app/javascript/packs/tpi.js
tsubik: Yeah, I feel guilty of adding jquery to the frontend bundle, but I guess it's still hard to avoid it as even many react libraries are just wrappers on jquery plugins. I added as I wanted `html` function to replaceHtml and eval scripts, maybe I could do it differently. Simple changing `innerHtml` of element is not enough.

app/javascript/tpi/controllers/index.js
rapzo: What for?

app/javascript/packs/tpi.js
rapzo: `even many react libraries are just wrappers on jquery plugins` only the bad, unmaintained ones.
But we can deal with it later on.

tsubik:
image

@simaob active level added. Of course, charts need some styling and axisX should not show that many ticks. When the chart was smaller not almost full screen it worked fine (only year was displayed on axis). Looks like this is some automatic thing.

app/javascript/tpi/controllers/index.js
tsubik: tooltips for sure.

PR #82, opened 2019-10-17
feature(frontend ui kit): configuration and scaffold

rapzo:

PR #93, opened 2019-11-06
Feature/rapzo frontend ui kit

tsubik:
Moved https://github.com/Vizzuality/laws_and_pathways/pull/82 here and fixed tests. Hopefully, nothing got broken. @rapzo can you check if I everything is here? The default route is still admin dashboard, TPI is under `/tpi`


simaob:
having a look too.

tsubik:
I removed create Publications migration here so, if anyone will have a problem please do `rails db:reset`

simaob:
I might've created these conflicts when I merge Marta's PR ... sorry

rapzo:
Fixed. Should be harmless.

PR #94, opened 2019-11-06
Update labels to say Laws instead of Legislations that doesn't exist

simaob:
Small PR to avoid distractions with non existent words as "Legislations"...


simaob:
Made a few more changes in titles and labels throughout the app... given the constant reminder that "Legislations" doesn't exist. Now I'm wondering if we should just rename the model to Law and the table to laws, and change all over the place, to be more consistent.

PR #95, opened 2019-11-06
Feature/ocd fixes

simaob:
In this PR:

- sorts tags in forms (though for keywords it doesn't really seem to work with the case sensitivity, but I lost hope now);
- sorts jurisdictions and geographies in forms;
- includes instrument types in instruments form selection;
- includes governance types in governances form selection;

I found out that formtastic hasn't been updated since 2017... and is a bit buggy and annoying.

### DIFF
--- a/app/models/governance.rb
+++ b/app/models/governance.rb
@@ -21,6 +21,6 @@ class Governance < ApplicationRecord
   validates_presence_of :name
 
   def self.to_select
-    self.pluck(Arel.sql("CONCAT('[', governance_types.name, '] ', governances.name)"), :id)
+    pluck(Arel.sql("CONCAT('[', governance_types.name, '] ', governances.name)"), :id)
   end
 end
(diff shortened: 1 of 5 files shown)


tsubik:
Not good, I thought that last changes were this year, but I see no release in couple years. What can we do if we use active admin?

simaob:
Damn, I pushed with broken specs apparently, sorry! Left the laptop in the office so can only fix tomorrow.

I was checking formtastic and they have commits early in the year, but no new releases. It kind of works, just needed to dig into the code to figure out some stuff. I should contribute some updates to the docs.

tsubik:
Right, that Tag select is broken. That must be select2 problem with some option.

tsubik:
I see another problem with that tags component. Items are doubled when we select more than 1. I thought that worked fine.

simaob:
What do you mean? doubled? I'll fix the rubocop complain and we can move on with this for now as you said.

tsubik:
image

Not sure what is causing this yet, but I'm going to fix that in separate PR.

simaob:
I saw that, but when I clicked an existing one, it would deselect it. So I thought that it was fine...

I've fixed the rubocop issues and the deprecation warnings

PR #96, opened 2019-11-07
fix(home): applying more style fixes and extra assets

rapzo:
- reworked stylesheets' directories


aabdaab:
Bring back `<%= javascript_pack_tag 'application' %>` to `tpi.html.erb` so the react components work 👍

app/assets/stylesheets/tpi/_navbar.scss
```diff
@@ -0,0 +1,27 @@
...
+    background-color: $dark;
+  }
+
+  .navbar-item,
+  .navbar-link {
+    padding: 0.5rem 1.25rem;
+  }
+
+  .navbar-dropdown {
+    @if $desktop {
```
aabdaab: haven't used @if conditionals in sass! is it like @media query?

app/assets/stylesheets/app/_sizes.scss
```diff
@@ -0,0 +1,10 @@
+// TEXT SIZE
+// $size-1: 3.5rem;
+// $size-2: 3rem;
+// $size-3: 2.5rem;
+// $size-4: 2rem;
+// $size-5: 1.5rem;
+// $size-6: 1rem;
```
aabdaab: are we gonna need them later?

app/assets/stylesheets/tpi/_navbar.scss
rapzo: Yap. But `$desktop` is defined in bulma though. We only configure the breakpoints variables in `application.scss` (for now, at least).

app/assets/stylesheets/app/_sizes.scss
rapzo: Bulma defines and uses the `$size-1` to `$size-7`. The commented values are the default values kept only for reference, since they match our specs.

aabdaab:
Thanks for rearranging the styles 👏 I'm gonna merge this but we need to reiterate on the home page (we should add PT task for that if there is none?)

PR #99, opened 2019-11-08
Feature/tpi company page styling

tsubik:
image

Still a lot of work to do, but it's a start.

I need to refactor those styles a little bit ;]


tsubik:
When navigation to Country page which is CCLOW page and clicking on back button styles are messed up. Probably picking some CCLOW styles? I didn't investigate that.

tsubik:
Actually, this one could be merged, and I could work on refactoring and styling more in a separate PR.

PR #101, opened 2019-11-11
Bubble Chart with real data

aabdaab:
Uff! It's here! After all, I removed `victory` library and created a custom chart with d3. The problem with `victory` was that out of the box the bubbles were distributed randomly so often they were overlapping each other. I played a bit with multiplying the `x`es and dividing `y`s to give the bubbles more space but that was too much dependent on the data.
My second approach involved using `d3-pack` but there were 2 issues with that:
1) The bubbles scaled depending on the number of bubbles per chart. That means that when there were ~10 bubbles per chart, the size of a _large_ bubble was as if it was a _small_ one. The legend would not be reliable in that case.
2) `d3-pack` "attracts" the bubbles towards the `y` axis and we wanted to distribute them more _horizontally_

In the end, I used `d3-force` that allows specifying to which axis the data should be aligned to.

8D2F7232-62FF-400D-B4CB-3596AA59E1BE

So, in this PR:
* Change `victory` to `d3`
* Bind real data with the chart
* Add tooltips to bubbles

**For a different PR**
* Legend
* Sector links
* Company links? (on bubble click should we be redirected to the company associated with the bubble?)


aabdaab:
I made the cell smaller so the bubble chart is now more concise:
image

aabdaab:
check now ☝️

app/javascript/components/SingleBubbleChart.js
```diff
@@ -0,0 +1,81 @@
+import React, { Fragment, PureComponent } from 'react';
+import { range } from 'd3-array';
+import { select } from 'd3-selection';
+import tip from 'd3-tip';
+import * as d3 from 'd3-force';
+
+const CELL_HEIGHT = 55; // same as cell height in bubble_chart.scss
+
+class SingleBubbleChart extends PureComponent {
```
rapzo: In case of doubt, function components are pure components.

PR #102, opened 2019-11-11
Feature/cclow climate targets ui

batraisk:


app/assets/stylesheets/cclow/_geography-page.scss
```diff
@@ -44,6 +44,12 @@
         margin-right: 1.5rem;
       }
     }
+
+    .highlight {
+      color: $red;
+      font-weight: bold;
```
rapzo: Not that it makes much difference but it is always best to use the bold font instead of stretching the regular font.

batraisk: Ok, I will fix it, thanks)

PR #112, opened 2019-11-15
Feature/change flag provider

simaob:
This PR updates the source of our flags, following from a recommendation by Fausto.
The updated flags are sourced from this repository: https://github.com/madebybowtie/FlagKit
I had to change the way we fetch the flags a little, because I get them directly from github using Oktokit.


tsubik:
I was about to say that fixed rubocop offences.

simaob:
I should code more, I forge these _details_

tsubik:
One more thing I have to push here.

tsubik:
EU and some other flag mapping. We also need International flag from somewhere and come up with code for this Supranational.

simaob:
Thanks for the fixes. Fausto wanted this up so he can check which flags are missing, including international.

PR #113, opened 2019-11-15
(spike): activity log for publishable resources

kowal:
# Summary

Add option to track changes in publishable models. Required cases:
```
- [X someone] published [record Y] on [date] - link to record;
- [X someone] archived [record Y] on [date] - link to record;
- [X someone] deleted [record Y] on [date];
- [X someone] unpublished [record Y] on [date] - link to record;
- [X someone] edited [record Y] on [date] - link to record;
```

### Details

Testing/comparing following solutions:

- `paper_trail`
  * separate `versions` table
  * association versioning - [via another gem](https://github.com/westonganger/paper_trail-association_tracking)
- `logidze`
  * only PostgreSQL 9.5+
  * requires SQL schema format
  * associations versioning - [experimental feature](https://github.com/palkan/logidze/wiki/Associations-versioning)
  * requires migration for every new tracked model
  * requires [custom code to build "change-sets" from records "versions"](https://github.com/Vizzuality/laws_and_pathways/pull/113/files#r347364488)
  * does not require joins with main table in case we're working only on given model (not our use-case right now)
- `public_activity`
  * separate `activities` table
  * by default doesn't store info which attributes have changed

Go to http://localhost:3000/admin/activity_log to see all libraries used (for different example models):

#### Testing
activity2

### DIFF
--- a/app/admin/activity_log.rb
+++ b/app/admin/activity_log.rb
@@ -1,0 +1,17 @@
+ActiveAdmin.register_page 'Activity Log' do
+  menu parent: 'Administration', priority: 3
+
+  content do
+    panel 'Activity Log' do
+      section 'Using paper_trail (Companies, Geographies)' do
+        table_for PaperTrail::Version.order('id desc').limit(20) do # Use PaperTrail::Version if this throws an error
+        column ("Item") { |v| v.item || '[Record deleted (TODO: find it by all_discarded scope)]' }
+        # column ("Item") { |v| link_to v.item, [:admin, v.item] } # Uncomment to display as link
+        column ("Type") { |v| v.item_type.underscore.humanize }
+        column ("Modified at") { |v| v.created_at.to_s :long }
+        column ("Admin") { |v| link_to AdminUser.find_by_email(v.whodunnit).email, [:admin, AdminUser.find_by_email(v.whodunnit)] }
+        end
+      end
+    end
+  end
+end
--- a/app/admin/activity_log.rb
+++ b/app/admin/activity_log.rb
@@ -4,48 +4,62 @@ ActiveAdmin.register_page 'Activity Log' do
   content do
     panel 'Activity Log (paper_trail), [Companies, Geo]' do
       section do
-        table_for PaperTrail::Version.order('id desc').limit(20) do
-          column ("Item") { |v| v.item || '[Record deleted (TODO: find it by all_discarded scope)]' }
-          column ("Type") { |v| v.item_type.underscore.humanize }
-          column ("Updated at") { |v| v.created_at.to_s :long }
-          column ("Whodunnit") { |v| v.whodunnit }
-          column ('Details') { |v| v.changeset }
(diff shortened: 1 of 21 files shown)


tsubik:
Need to think about it. PaperTrail takes a lot of space, need to be disabled for imports for sure. It's also possible to just pick attributes that we monitor, but that's onlu for Active Models. I probably would dismiss logidze and database triggers and think if we need to track some not active record related events in the future (user logged in, payments). In one project I remember I used papertrail but only to track record changes and for activityLog just simple `Activity` model with closed list of all possible activites. All interaction were through `Interactors` objects without using callbacks which means that wasn't plug-and-play solution like PaperTrail but I had full control over it.

simaob:
Have you checked this option: https://github.com/chaps-io/public_activity, first result on _my_ Google when I search for tracking this on ActiveAdmin.

app/admin/activity_log.rb
```diff
@@ -0,0 +1,67 @@
...
+        changesets_logidze = klasses.flat_map do |klass|
+          klass.where.not(log_data: nil).order(updated_at: :desc).flat_map do |record|
+            record_updates = record.log_data.versions.each_cons(2).map do |before, after|
+              record.at(time: after.time).diff_from(time: before.time)
+            end
+
+            record_updates.reverse.compact.map do |record_update|
+              single_record_changes = record_update['changes']
+                .except('updated_at')
+                .map { |name, changes| {name => [changes['old'], changes['new']]} }
```
kowal: With `logidze` you need built record "changesets" (`record_updates` in code above) for yourself - it is not a lot of code but it is kinda surprising because library is advertised as a tool for "logging DB records **changes**".

kowal:
> Have you checked this option: https://github.com/chaps-io/public_activity, first result on _my_ Google when I search for tracking this on ActiveAdmin.

@simaob in `public_activity` you also need custom solution to store actual changes (via parameters).

simaob:
> > Have you checked this option: https://github.com/chaps-io/public_activity, first result on _my_ Google when I search for tracking this on ActiveAdmin.
> 
> @simaob in `public_activity` you also need custom solution to store actual changes (via parameters).

Ohhhh, but I wasn't thinking of storing the actual changes. Just that something changed. "This record was edited", no need to say "This record was edited and the changes were". They can use the comments to inform each other of what needs reviewing!

kowal:
> 
> Ohhhh, but I wasn't thinking of storing the actual changes. Just that something changed. "This record was edited", no need to say "This record was edited and the changes were". They can use the comments to inform each other of what needs reviewing!

@simaob @tsubik ok, for some reason I thought that we need those details :sweat_smile: 

I opened another PR with `public_activity` gem added - the thing is, in order to map updates for `visibility_status` we need some custom mapping anyway - the way I implement it, makes it really easy to actually store those change-sets, so we're in good place if requirement changes at some point.

I also pushed those changes here, for easy comparison of all 3 libraries - might be useful in the future ;)

PR #115, opened 2019-11-17
Feature/tpi styling vol2

tsubik:
- Adding tooltips to company page (I think we may consider using some tooltip library instead, like tippy.js)
- Fixing CP emission chart data (sometimes year 2013, 2014 were at the end, how knows why, problem is gone when providing year as an integer not a string)
- couple small fixes with styling, company properties, trend badges (should be on top of each other when star + up trend for example), adding some base styles for inputs, using it for select.


simaob:
we can leave the tooltip changes to a different PR.

tsubik:
Yeah, other PR :+1:

PR #119, opened 2019-11-18
add public-activity for tracking updates in publishable records

kowal:
#### Summary

* Adds public_activity page
* tracks `Company, Geography, Legislation, Litigation, Target` resources
* recognizes changes to `visibility_status`, deletes and all other "edits"
* only single updates for now (bulk updates skips callbacks)

act200

### DIFF
--- a/app/models/concerns/public_activity_trackable.rb
+++ b/app/models/concerns/public_activity_trackable.rb
@@ -24,9 +24,7 @@ module PublicActivityTrackable
   end
 
   def owner
-    return unless respond_to?(:updated_by)
-
-    updated_by
+    ::Current.admin_user
   end
 
   def activity_key_for_last_update
--- a/app/models/concerns/public_activity_trackable.rb
+++ b/app/models/concerns/public_activity_trackable.rb
@@ -1,0 +1,41 @@
+# Tracks updates of:
+# - common attributes updates (visibility_status, discarded_at)
+# - deletes
+#
+# Required to be implemmented by the model
+# - create_activity (from PublicActivity::Common)
+# - updated_by (from UserTrackable)
+#
+module PublicActivityTrackable
+  extend ActiveSupport::Concern
+
+  included do
+    around_update :track_updates_activity
+  end
+
+  protected
+
+  def track_updates_activity
+    yield
+
+    return unless errors.blank? && previous_changes.present?
+
+    create_activity activity_key_for_last_update, owner: owner
+  end
+
+  def owner
+    return unless respond_to?(:updated_by)
+
+    updated_by
+  end
+
+  def activity_key_for_last_update
+    if previous_changes['visibility_status']
+      previous_changes['visibility_status'].last
+    elsif previous_changes['discarded_at']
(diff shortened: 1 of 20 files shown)


tsubik:
What should we do with bulk upload and bulk actions? Just one activity (disable that callback), or bunch of activities for each update. Something to consider.

simaob:
So on the other PR I mentioned a single activity, but I guess it will depend on how much they use it, maybe for now I'd keep with tracking all of the activities individually and if see how it looks. Otherwise we might need to tweak the display so that they can see exactly what was affected.

simaob:
One thing to add is the link to the resource, so that it's easy to go to that page and check. Other than that, this is what I wanted =D We might want to add pagination to the page, in case this grows quite large.

tsubik:
This page could be a standard active admin index page, right? With filters, pagination, everything. Last activities could be another section that we will add to the dashboard.

kowal:
I'll get back to this tomorrow. 
@tsubik good point with standard active admin index page, will check that.

kowal:
> What should we do with bulk upload and bulk actions? Just one activity (disable that callback), or bunch of activities for each update. Something to consider.

Bulk updates are currently implemented using `update_all` which skips validations & callbacks, so we won't record any activity automatically. We should get more info of what we really want here and update batch actions classes accordingly but maybe let's do it i another story.

kowal:
@simaob @tsubik please review.

I won't have much time during upcoming week, so if you find some small things to fix, feel free to take this over! thanks ;)

kowal:
rebase after https://github.com/Vizzuality/laws_and_pathways/pull/131 was merged

tsubik:
CSV import will generate as many activities as there were updates, maybe to be consistent bulk update should also do the same? I'm not sure.

app/models/concerns/public_activity_trackable.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  protected
+
+  def track_updates_activity
+    yield
+
+    return unless errors.blank? && previous_changes.present?
+
+    create_activity activity_key_for_last_update, owner: owner
```
kowal: If we want to manually create activities (for bulk updates, CSV imports) we could extract from here something like `#create_activity_with_current_changes`:

```ruby
module PublicActivityTrackable
  # ..
  def create_activity_with_current_changes!
    # ..
    create_activity activity_key_for_last_update, owner: owner
  end

  protected

  def track_updates_activity
    yield
    create_activity_with_current_changes!
  end
end
```

then for mentioned cases:

```ruby
company.update_attributes()
company.create_activity_with_current_changes!
```

tsubik: I think we could just start without using `update_all` in batch actions.

tsubik: Just `update` and everything should work.

tsubik: I've done in this way. Looks like it works as it should now I think

PR #121, opened 2019-11-19
Feature/small back end tasks

simaob:
- Adds batch actions for Litigations and Targets (delete, archive, publish);
- Adds publish batch action to Laws;

### DIFF
--- a/app/admin/litigations.rb
+++ b/app/admin/litigations.rb
@@ -112,6 +112,20 @@ ActiveAdmin.register Litigation do
 
   form partial: 'form'
 
+  batch_action :publish,
+               priority: 1,
+               if: proc { current_scope&.name != 'Published' } do |ids|
+    publish_command = ::Command::Batch::Publish.new(batch_action_collection, ids)
+
+    message = if publish_command.call
+                {notice: "Successfully published #{ids.count} Litigations"}
+              else
+                {alert: 'Could not publish selected Litigations'}
+              end
+
+    redirect_to collection_path(scope: 'published'), message
+  end
+
   batch_action :archive,
                priority: 1,
                if: proc { current_scope&.name != 'Archived' } do |ids|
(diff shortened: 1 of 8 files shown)


app/admin/litigations.rb
```diff
@@ -113,6 +112,46 @@
 
   form partial: 'form'
 
+  batch_action :publish,
```
tsubik: See a potential here to refactor those to some kind of DSL extensions that we have. But could be done in another PR.

app/commands/command/batch/publish.rb
```diff
@@ -0,0 +1,16 @@
...
+    class Publish
+      def initialize(collection, ids)
+        @collection = collection
+        @ids = ids
+      end
+
+      def call
+        @collection
+          .where(id: [*@ids])
+          .update_all(visibility_status: 'published')
```
tsubik: `update_all` won't call any callbacks, and when we merge public_activity PR we either need to add one activity like "George published x Litigations" and put ids in the comment. Or fire `@collection.where(id: [*ids]).update(visibility_status: 'published')`.

tsubik: Which will generate bunch of activites. Not sure what is better, I think just one batch activity seems fine.

spec/factories/litigations.rb
```diff
@@ -32,7 +32,7 @@
     citation_reference_number { 'SFKD777FDK77' }
     summary { 'Summary Lorem ipsum dolor dalej nie pamietam' }
     core_objective { 'Core objectives Lorem ipsumumum' }
-    visibility_status { Litigation::VISIBILITY.sample }
+    visibility_status { Litigation::VISIBILITY.first }
```
tsubik: right, maybe better just pick 'published'

app/commands/command/batch/publish.rb
simaob: yeah, I think one batch activity would work well, and keep it saner. I'll update it.

simaob: actually, let's leave it like this and then we address capturing these on the activities.

PR #122, opened 2019-11-19
setting up production deploy files for cap

simaob:
Would be great to get your help with this, @tiagojsag.

Currently we are using `laws-pathways` as a staging site and now I've got a new subdomain of `laws-pathways-staging.vizzuality.com`, should we start using the current site and database as production and create the apache2 conf and database for staging? We can purge the current database and load only the relevant data. What do you think is easiest?

- [ ] Setup apache2 conf;
- [ ] Create new database for new environment;
- [ ] Setup SSL certificate;

### DIFF
--- a/config/deploy/production.rb
+++ b/config/deploy/production.rb
@@ -3,7 +3,6 @@ set :ssh_options, forward_agent: true
 
 set :branch, 'master'
 
-# Default deploy_to directory is /var/www/my_app_name
-set :deploy_to, '/var/www/laws-pathways'
+set :deploy_to, '/var/www/laws-pathways-production'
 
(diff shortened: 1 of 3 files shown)


config/deploy/production.rb
```diff
@@ -1,55 +1,9 @@
...
-#   user: "user_name",
-#   roles: %w{web app},
-#   ssh_options: {
-#     user: "user_name", # overrides user setting above
-#     keys: %w(/home/user_name/.ssh/id_rsa),
-#     forward_agent: false,
-#     auth_methods: %w(publickey password)
-#     # password: "please use keys"
-#   }
+# append :linked_files, 'config/secrets/google-credentials-staging.json'
```
simaob: Also help with this :D

PR #123, opened 2019-11-20
Feature/tweak litigations

simaob:
This PR makes some updates to the litigations table based on conversations here in London:

- renames `core_objective` to `at_issue`;
- updates `document_types`, will need an updated document for the importer of this, that I will deal with later;
- adds `responses` as tags to litigations and legislations (this is a new type of tag to be used now);
- adds parent_legislation to legislations;
- small tweaks to the order of display of attributes.


simaob:
need to change some stuff, WIP flag added.

simaob:
PR updated and ready now. I have force-pushed a couple of times. Description is up to date so do read it, as I end up doing a few more things in this PR.

simaob:
Hey Tomasz, hold on! I think I added an extra field by accident to the legislations model. Will check it and try to fix, I think it's the migration I created that is broken.

tsubik:
Ok

simaob:
done! I hope :D

db/migrate/20191120152618_add_parent_to_legislations.rb
```diff
@@ -0,0 +1,5 @@
+class AddParentToLegislations < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :legislations, :parent, null: true
```
tsubik: I think it should be
```
add_reference :legislations, :parent, foreign_key: { to_table: :legislations }, index: true, null: true
```

not sure if we need index

tsubik: but we need foreign key

simaob: I'm doing this in a middle of another chat now, so sorry for this. Maybe I should squash some of these commits to avoid messes for other people.

tsubik: no need to squash I think

tsubik:
Just one thing left, `rails db:reset` does not work, because `case` is no longer a valid litigation document_type. We need to update litigation.csv with dev data.

simaob:
Yes, that's true. I was going to deal with that on a different PR. but yes it's outstanding!
Shall I fix it here or can we live this, what'ya think?

tsubik:
I will push a quick fix here.

PR #124, opened 2019-11-20
Rails6

tsubik:
We may need this sooner than later. Would be good to separate prod and staging credentials which Rails 6 supports natively, also ActiveStorage is more mature there.

Still have some failing test, and need to manually test it, but looks so far.


simaob:
Looking good locally! =)

tsubik:
I think it should be fine to merge it and test on staging.

PR #127, opened 2019-11-21
Feature/cclow latest additions

batraisk:


app/views/cclow/home/_latest_additions.html.erb
```diff
@@ -0,0 +1,24 @@
+<div class="slider-container">
+  <div class="slider" data-slick='{"slidesToShow": 3, "slidesToScroll": 3}'>
+    <% @latest_additions.each do |item| %>
+      <div class="item-container">
+        <div class="item">
+          <div class="additions-kind"><%= item[:kind] %></div>
+          <div class="title"><a href="#"><%= item[:title] %></a></div>
```
simaob: When clicking on the title it should lead us either to the legislation show page or the litigation show page. Could you hook that up in this PR?

batraisk: yes, of course, I will do it now

app/views/cclow/home/_latest_additions.html.erb
```diff
@@ -0,0 +1,24 @@
...
+            <div><%= item[:jurisdiction] %>  |  <%= item[:addition_type] %>  |  <%= item[:date_passed] %></div>
+          </div>
+        </div>
+      </div>
+    <% end %>
+  </div>
+</div>
+
+
+<%= javascript_tag do %>
```
aabdaab: For consistency and readability reasons I'd move html + js from this file to react component and render it here like all the other ones:
```
<%= react_component("YourSliderComponeny", { your props }) %>
```
you also need to add `onClick` functionality to the slider box

app/views/cclow/home/_latest_additions.html.erb
```diff
@@ -0,0 +1,24 @@
+<div class="slider-container">
+  <div class="slider" data-slick='{"slidesToShow": 3, "slidesToScroll": 3}'>
+    <% @latest_additions.each do |item| %>
+      <div class="item-container">
+        <div class="item">
+          <div class="additions-kind"><%= item[:kind] %></div>
+          <div class="title"><a href="#"><%= item[:title] %></a></div>
+          <div class="meta">
+            <img class="flag" src=<%= asset_path "flags/#{item[:iso]}.svg"%> alt="">
+            <div><%= item[:jurisdiction] %>  |  <%= item[:addition_type] %>  |  <%= item[:date_passed] %></div>
```
simaob: Maybe it's because of the data that we have but at the moment because we don't have some of these attributes, sometimes we are displaying unnecessary `|` . So we should check for the presence of the data and then join with pipes.

app/assets/stylesheets/cclow/_latest-additions-slider.scss
```diff
@@ -0,0 +1,78 @@
+@import "variables";
+
+.slider-container {
+  .slick-slide {
+    padding: 0 10px;
+  }
+
+  .slider {
+    margin: 0 -10px;
```
aabdaab: does this `10px` depend on the 10px in `slick-slide` class? meaning: if I change it to -15 for example, will the slick-slide need to be changed too? if so, please move it to a variable

app/services/api/latest_additions.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    def call
+      (serialize_litigations + serialize_legislations).sort_by { |item| - item[:created_at].to_time.to_i }
+    end
+
+    private
+
+    def serialize_litigations
+      Litigation.last(@count).map do |item|
```
simaob: I think for now this logic works. we need to see once we have proper data, because we might want to filter on specific types of events instead of getting just the latest laws and litigations. but for now should be fine!

app/services/api/latest_additions.rb
```diff
@@ -0,0 +1,40 @@
+module Api
+  class LatestAdditions
+    def initialize(count)
+      @count = count
+    end
+
+    def call
+      (serialize_litigations + serialize_legislations).sort_by { |item| - item[:created_at].to_time.to_i }
```
aabdaab: it's just a question, maybe it's supposed to be this way... But I thought that the latest additions don't mean latest in the terms of adding them to the DB (`created_at`) but more in a sense of `date_passed` in Legislation

simaob: Yes, we don't have the right data in, so I'd leave as this for now and then adjust once we got some importers in.

app/views/cclow/home/_latest_additions.html.erb
tsubik: Hmm, not sure if we need react for that simple carousel.

tsubik: it's a slick slider.

app/services/api/latest_additions.rb
tsubik: at least those should be published ones. `Litigation.published` but that's for everything we need on frontend.

app/views/cclow/home/_latest_additions.html.erb
tsubik: Could be also done with stimulus `slider_controller`

```
 <div class="slider" data-controller='slider-controller' data-slick='{"slidesToShow": 3, "slidesToScroll": 3}'>
```

In `shared/controllers/slider_controller.js`
```
import { Controller } from 'stimulus';

export default class extends Controller {
  connect() {
    this.element.slick({
      prevArrow: '<button type="button" class="slick-prev"><i class="fa fa-1x fa-arrow-left"></i></button>',
      nextArrow: '<button type="button" class="slick-next"><i class="fa fa-1x fa-arrow-right"></i></button>'
    });
  }
  
  disconnect() {
     this.element.unslick();
  }
}
```
Something like this.

tsubik: Something similar I would do with tooltips.

app/views/cclow/home/_latest_additions.html.erb
tsubik: could be like `item.values_at(:jurisdiction, :addition_type, :date_passed).compact.join(' | ').html_safe`

tsubik: Jurisdiction should be a link actually

app/services/api/latest_additions.rb
```diff
@@ -0,0 +1,40 @@
+module Api
+  class LatestAdditions
```
tsubik: Simple test, just checking if the right data is returned would be nice too :+1:

app/views/cclow/home/_latest_additions.html.erb
rapzo: There's also the testimonials (quotes?) component that should be fairly simple to adapt to this scenario.

app/views/cclow/home/_latest_additions.html.erb
rapzo: Just an aside: it is valid to use anchors (<a></a>) without href. Having the `#` with no proper `prevendDefault()` behaviour may cause undesirable flickers.
Or there's always the `<button class="link">` that i think it will look the same, but i'd stick to anchors with no href.

spec/services/api/latest_additions_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+RSpec.describe Api::LatestAdditions do
```
tsubik: :+1:

app/javascript/shared/controllers/slider_controller.js
```diff
@@ -0,0 +1,14 @@
+import { Controller } from 'stimulus';
+
+export default class extends Controller {
```
tsubik: :+1:

PR #129, opened 2019-11-21
feature(tpi: publications): add publications mock pages

rapzo:


app/assets/stylesheets/publications.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the publications controller here.
```
tsubik: to be removed I guess.

rapzo: Yes, sorry.

PR #130, opened 2019-11-21
Fuzzy search/dropdown for Laws

aabdaab:
This PR adds a global search bar to the `CCLOW` project. 
Designs: https://projects.invisionapp.com/d/main#/console/18784734/391067162/preview

screencast 2019-11-21 15-53-04

* It's added to every endpoint on `/cclow`
* As long as there are no attributes for laws/litigations/targets from which we could determine the recent additions, these items have disclaimers `todo`
* As a point of reference for "recent" I set `1.year.ago` - but that can be of course adjusted accordingly.
* Added shared `SearchController` that is then included in every cclow controller (search dropdown needs specific data) - there shouldn't be any attribute clashes, I added a `search_` prefix to all of the class variables
* `Enter` functionality wasn't added deliberately - it's not clear what would be selected in that case
* Click outside of the search dropdown and ESC key on the keyboard close the search bar
* Handle query params in the controllers for redirecting to the search results from the frontend

**The logic goes as follows:**
* If no search input
   - show all of the options
   - if there was a search before **and we clicked the result** - show the Last Search section
* If there is a search input
   and there are some matching results
      - only show the categories with matching results
   and there are no matching results
      - show the "no results" disclaimer
      - show all options below


tsubik:
Looks awesome. 

Just I think this one needs to work directly with some API endpoint, and I'm thinking we should drop that xx Results feature for full-text search (as it should look in title, description, and probably in other attributes). We are going to overheat our CPUs with the unnecessary search of everything every time the user type anything.

tsubik:
Actually, nvm, we cannot drop that search everywhere without totally changing this component. We will have to resort to good debouncing and figure out something on the server to not kill it that quickly :P

Full-text search should be done at least on: title, description, other text fields, all tags and keywords, maybe connected entities names?

This is for another story as not a small task.

simaob:
This is awesome, Marta!! Good job! =)
I think this PR can be merged and the API changes and full text search improvements can come in separately, what you do guys think?

aabdaab:
hey, Tomek, you might be onto something; I didn't have any issues so far on my machine but I turned on the CPU throttling and there's some bottleneck when typing the input. I'm gonna experiment with the `debounce` as you suggested and also selecting only the attributes that we need on the frontend when pulling them from DB in the controllers cause the fuse js is likely going through all of them, even on those we don't really need to account for. I'll try in this PR so don't merge ❌ ✌️

aabdaab:
@simaob I'll take like 2 hours, if it's gonna take longer, then let's do a different PR ✌️

simaob:
That's cool, we can wait!=)

aabdaab:
Ok, there are two things I changed:
* I queried from DB only the text fields in the `search_controller` - the full-text search only needs these to go through
* Debounced the searching - once the user stops typing only then it performs the search (it actually sets the input value into the React state and upon that the search is triggered (`useMemo` hooks)

aabdaab:
Ready to merge if you're happy ☝️

PR #131, opened 2019-11-21
TPI Production Ready Seeds

simaob:
This PR updates the TPI scripts and data files with the most up to date information, which we can use to seed production.

Main changes are:

- we are using the existing TPI ids, instead of creating new ones each time, making it easy to link data;
- isin stopped being an unique value, because they had to duplicate some companies that are assessed in different sectors;
- updated companies script to import the new fields that were added recently;


simaob:
oh I'm also importing Instrument Types, Instruments and LawSectors in this PR, did it directly in the seeds.rb file.

simaob:
Fixed specs and rubocop and added company_id to MQ and CP assessments download, as this is now used to import the data.

app/models/company.rb
```diff
@@ -46,7 +46,7 @@ class Company < ApplicationRecord
 
   validates :ca100, inclusion: {in: [true, false]}
   validates_presence_of :name, :slug, :isin, :market_cap_group
-  validates_uniqueness_of :slug, :isin, :name
```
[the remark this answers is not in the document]
simaob: yeah, I missed that too 😅

simaob:
Thanks for the feedback, I'll check it again later today, including the override small refactor.
In regards to the MQ version, I think I can pass it in the seeds, but then maybe we increment it on importing, so that it's sequential and doesn't need to be provided by the file each time. What do you think?

simaob:
Refactored to have override_id as an option, and added methodology_version, added it to the importing files. The dates are so different that was hard to extract from that...

Looking now at the CP Assessments assessment_date

simaob:
the cp-assessments file had the wrong format on the assessment date... fixed now. Should be good to go, unless there's feedback on my refactor! =D

app/services/csv_import/base_importer.rb
```diff
@@ -60,6 +65,12 @@ def csv_converters
 
     private
 
+    def reset_id_seq
```
tsubik: :+1:

app/services/csv_import/mq_assessments.rb
```diff
@@ -52,7 +52,8 @@ def assessment_attributes(row)
         publication_date: publication_date(row),
         level: row[:level],
         notes: row[:notes],
-        questions: get_questions(row)
+        questions: get_questions(row),
+        methodology_version: row[:methodology_version]
```
tsubik: :+1:

PR #133, opened 2019-11-25
Release/tpi production seeds

simaob:
We are moving towards release! =)

Besides bringing a lot of visual improvements to both TPI and CCLoW, this PR brings also changes to the backend and the TPI seeds of data to add to Production.


tsubik:
Should be fine without checking, if fine on staging ;]

PR #134, opened 2019-11-25
Add static pages management and dynamic routes

aabdaab:
This PR adds models for static pages (think the content of `About` tab):
* Page - that will hold its title, slug and short description
* Content - content of the page, page can have many contents - mostly did this because of `Supporters` page that has two different sections with related logos. Content will have text and images.
* DynamicRouter - for adding dynamic routes when new page is added.

I added the `Pages` section in ActiveAdmin, under TPI:
Screenshot 2019-11-25 at 13 30 16

I tried to follow the example of the `Litigations` form that has associations with `litigation_sides` - I wanted to do an equivalent for Page and its Contents.

When you add a new Page, you can go to `/tpi/:slug` to see that the route was created.

Before testing:
- run `rake db:migrate`


db/schema.rb
```diff
@@ -136,6 +145,11 @@
     t.index ["documentable_type", "documentable_id"], name: "index_documents_on_documentable_type_and_documentable_id"
   end
 
+  create_table "dynamic_routers", force: :cascade do |t|
```
tsubik: Why do we need this table?

app/models/dynamic_router.rb
```diff
@@ -0,0 +1,14 @@
+class DynamicRouter < ApplicationRecord
```
tsubik: That could be a service?

tsubik: Or we need this to be persisted model in DB?

aabdaab:
Fixed saving the contents ☝️Also noticed that tests are not failing

app/models/page.rb
```diff
@@ -0,0 +1,16 @@
...
+  has_many :contents
+
+  validates :slug, uniqueness: true, presence: true
+  validates :title, presence: true
+
+  with_options allow_destroy: true, reject_if: :all_blank do
+    accepts_nested_attributes_for :contents
+  end
+
+  after_save :reload_routes
```
tsubik: I would do this in `after_commit`

app/models/dynamic_router.rb
aabdaab: I guess it could, it'd be more verbose.. let me check

app/admin/page.rb
```diff
@@ -3,7 +3,7 @@
 
   menu priority: 6, parent: 'TPI'
 
-  permit_params :title, :slug, :description, contents_attributes: permit_params_for(:contents), content_ids: []
+  permit_params :title, :slug, :description, :active_tab, contents_attributes: [:title, :text], content_ids: []
```
tsubik: I don't think we need :active_tab here

app/models/dynamic_router.rb
tsubik: I mean it could be just moved to `services` and not inherit after `ApplicationRecord` if we don't need the model and the table

aabdaab:
Addressed the comments ☝️ 
I'm checking now why the tests are failing

aabdaab:
* Fixed tests
* Added uploading images
* Fixed issue with duplicated content records on the edit form - due to lack of `:id` in permitted params for content

There's still an issue with rendering the already uploaded images (`image_tag` apparently not working with ActiveStorage this way) and uploading more images in the edit form once we already uploaded some - but I think it can be handled in the next PRs?

tsubik:
Great. I can take a look at those images. Checking this now.

app/helpers/contents_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ContentsHelper
```
tsubik: we can remove that

app/helpers/pages_helper.rb
```diff
@@ -0,0 +1,2 @@
+module PagesHelper
```
tsubik: this too

aabdaab: 👍

spec/factories/dynamic_routers.rb
```diff
@@ -0,0 +1,4 @@
+FactoryBot.define do
```
tsubik: this too, to remove

spec/decorators/content_decorator_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
```
tsubik: hmm, not sure if we need to keep empty tests

app/assets/stylesheets/contents.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the Contents controller here.
```
tsubik: to remove

app/controllers/tpi/contents_controller.rb
```diff
@@ -0,0 +1,4 @@
+module TPI
```
tsubik: we dont need that too.

config/routes.rb
```diff
@@ -15,7 +15,7 @@
         get :emissions_chart_data
       end
     end
-    
+
```
simaob: just trailing spaces removal!

tsubik:
@aabdaab you could use decorators and for all html content use `html_safe` take a look at `TargetDecorator` for example.

aabdaab:
I added a few changes to this PR:
* new `Image` model that holds a `logo` and a `link`
* added a rake task `rake static_pages:generate` that's gonna scaffold some known Pages and its Contents - you can then change it in Active Admin, add logos to it, links etc.

image

aabdaab:
I added a rake task that creates some of the `Page` records beforehand and changed the `DynamicRouter`'s condition to pick up the routes and now the tests are failing again (on Travis, locally they're passing). I'm gonna look into this in more depth tomorrow.

PR #135, opened 2019-11-25
Tpi/frontend tweaks

tsubik:
Couple tweaks

- Adding sector to AA company index page
- changing colors and a few styles for pie charts
- missing companies number for sector's pie chart of levels
- changing colors for stacked bar chart, sectors cp benchmarks
- changing colors, xAxis interval, yAxis title placement for MQ assessment company level chart
- add Sedol to frontend (companies have 0 as this number, could be bug with import script @simaob)

Tweaking charts is a slow process, highcharts are the most customizable charts I've ever seen, milion options, just finding the one takes time.

CP assessment chart left intact, I couldn't find a good way to make those areas look as on designs yet.

### DIFF
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -16,6 +16,7 @@ ActiveAdmin.register Company do
   filter :isin_contains, label: 'ISIN'
   filter :name_contains, label: 'Name'
   filter :geography
+  filter :sector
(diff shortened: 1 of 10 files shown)


simaob:
so apparently SEDOL should be a string. fixing it and will push into this PR.

PR #136, opened 2019-11-26
Feature/cclow events timeline

batraisk:


app/assets/stylesheets/cclow/_events_timeline.scss
```diff
@@ -1,10 +1,13 @@
 @import 'variables';
 
+$arrow-btn-width: 48px;
+
 .events-container {
   display: flex;
   position: relative;
   overflow-y: hidden;
```
tsubik: I think it should be just `overflow: hidden` otherwise I have horizontal scrollbar, we should use navigation arrrows only.

app/javascript/components/EventsTimeline.js
```diff
@@ -0,0 +1,108 @@
...
+            {events.map((event, i) => (
+              <div key={i} className="time-point">
+                <div className="event-title">{ event.title }</div>
+                <div className="point"></div>
+                <div className="date">{ moment(event.date).format('MMMM Y') }</div>
+              </div>
+            ))}
+          </div>
+        </div>
+        <div className="arrow-right" onClick={() => el.current.scrollBy(40, 0)}>
```
tsubik: I would say it should scroll by the width of one event, actually more than 40px. Can we also making smoothscroll?
example 
```
scrollBy({
  top: 100,
  left: 100,
  behavior: 'smooth'
});
```
should work in most of the browsers.

tsubik:
Sorry about that spamming of request changes, I had too many tabs with this review open ;]

tsubik:
image

We should also drop this tag behavior. I think it should be like if you select one then the name of that selected event should appear, if you select more, then "2 selected", "3 selected". cc @simaob

app/javascript/components/EventsTimeline.js
```diff
@@ -0,0 +1,108 @@
...
+        <div className="arrow-left" onClick={() => {el.current.scrollBy(-40, 0)}}>
+          <i className="fa fa-arrow-left"></i>
+        </div>
+        <div ref={ el } className="events-container">
+          <div className="timeline">
+            {events.map((event, i) => (
+              <div key={i} className="time-point">
+                <div className="event-title">{ event.title }</div>
+                <div className="point"></div>
+                <div className="date">{ moment(event.date).format('MMMM Y') }</div>
```
rapzo: We shouldn't add `moment` because of this. It's a huge load. [date-fns](https://date-fns.org/) is a better option because it was designed with bundle size in mind (tree-shaking and dead code elimination work while in moment is doesn't).

app/javascript/components/EventsTimeline.js
```diff
@@ -0,0 +1,108 @@
...
+    </components.Option>
+  </div>
+);
+
+const EventsTimeline = ({ events, options, isFiltered = false }) => {
+  const [currentTypes, setCurrentTypes] = useState(false);
+  const el = React.createRef();
+    if (isFiltered && (currentTypes || []).length !== 0) {
+      const types = currentTypes.map(e => e.value);
+      events = events.filter(e => types.includes(e.event_type));
```
rapzo: Changing a prop? Is this a good idea?

app/javascript/components/EventsTimeline.js
```diff
@@ -0,0 +1,108 @@
...
+      {props.isSelected && <div className="checked select-checkbox"><i className="fa fa-check"></i></div>}
+      {!props.isSelected && <div className="unchecked select-checkbox"></div>}
+      <label>{props.label}</label>
+    </components.Option>
+  </div>
+);
+
+const EventsTimeline = ({ events, options, isFiltered = false }) => {
+  const [currentTypes, setCurrentTypes] = useState(false);
+  const el = React.createRef();
```
rapzo: A better name here would be nice. Took a bit to figure out which element el was.

tsubik:
Checking :eyes:

tsubik:
I just pushed one small change. Otherwise Looks great! Great work :)

PR #137, opened 2019-11-27
Feat: world map component

sorodrigo:
- Created World Map component using the Peters Gall projection
- Topojson from react-simple-maps github repo, we can download it if it matches our needs or we can use a different.
- Created a couple of fake layers with both ramps.
- For the ramps I'm using 2 different scales: square root for size, quantize for colors.
- Added zoom buttons, missing correct icons cause aint nobody got time for that.

laws-map


app/javascript/components/world-map/WorldMap.js
```diff
@@ -0,0 +1,136 @@
...
+  Geographies
+} from "react-simple-maps";
+import { geoCylindricalEqualArea } from "d3-geo-projection";
+import reducer, { initialState } from "./world-map.reducer";
+import { useMarkers } from "./world-map.hooks";
+import MapBubble from "./MapBubble";
+
+const geoUrl =
+  "https://raw.githubusercontent.com/zcreativelabs/react-simple-maps/master/topojson-maps/world-110m.json";
+const PetersGall = geoCylindricalEqualArea().parallel(45);
```
aabdaab: The map is a bit compressed, we need to try a different value for the parallel function, maybe `30`?

sorodrigo: Not possible check with fausto

tsubik: Right, looks weird, but if the client wants that projection then :man_shrugging:

simaob:
Tomasz or Marta could you check the conflicts and merge this? Rodrigo is now on leave!

PR #138, opened 2019-11-27
improve(tpi): pie charts design

rapzo:


tsubik:
If you rebase with current develop, then CI should pass.

rapzo:
Oh, that one, didn't go through. I applied proper column and offset usage according to specs in the other one. That is only missing the same container markup.
Will take care of it in a sec!

PR #140, opened 2019-11-28
Style static pages

aabdaab:
This PR styles the content of static pages managed through the admin panel.
It also fixes a nasty bug with saving multiple records for a nested association; there was an issue on Page creation that when you added a Content and then several images for that Content, only the last image was saved into the database.
The issue was caused by broken uniqueness of image template; when you were adding a new Content the `addRecord` method was replacing all the occurrences of `NEW_RECORD` key and for the image template as well, with the same Date. As a result, the keys weren't unique and only the last image was being saved.
It was fixed by adding a different key for the Image form `NEW_IMAGE_RECORD` and calling a different method `addNestedRecord` which then replaces that key with a Date which is now unique.

To test the whole workflow:
* run a rake task `rake static_pages:generate` that scaffolds the static pages - without the content though,
* go to `/admin/pages` and fill out the pages with some data,
* go to `tpi/:slug` and check the static page there

Below you can find a description of what's what in the Form in the admin panel. If it's a simple page, like `Methodology` which doesn't have any images on its page ([design](https://projects.invisionapp.com/d/main#/console/18479580/384135994/preview)), you don't need to add `Content` - you can add all the body in the `description` field of the `Page`.

TPIStaticPages

image

### DIFF
--- a/lib/tasks/static_pages.rake
+++ b/lib/tasks/static_pages.rake
@@ -4,21 +4,21 @@ namespace :static_pages do
     MENU_HEADERS = OpenStruct.new(Hash[Page::MENU_HEADERS.map { |el| [el, el] }])
 
     pages = [
-      ['overview', 'Overview of the TPI', MENU_HEADERS.about],
-      ['strategic-relationships', 'Strategic Relationships', MENU_HEADERS.about],
-      ['technical-advisory-group', 'Technical Advisory Group', MENU_HEADERS.about],
-      ['supporters', 'Supporters', MENU_HEADERS.about],
-      ['investors', 'How Investors can use TPI', MENU_HEADERS.about],
-      ['endorsements', 'Endorsements', MENU_HEADERS.about],
-      ['team', 'Team', MENU_HEADERS.about],
-      ['faq', 'FAQ', MENU_HEADERS.about],
-      ['methodology', 'Methodology', MENU_HEADERS.tpi_tool],
-      ['data-background', 'Data Background', MENU_HEADERS.tpi_tool]
+      ['Overview of the TPI', MENU_HEADERS.about],
+      ['Strategic Relationships', MENU_HEADERS.about],
+      ['Technical Advisory Group', MENU_HEADERS.about],
+      ['Supporters', MENU_HEADERS.about],
+      ['How Investors can use TPI', MENU_HEADERS.about],
+      ['Endorsements', MENU_HEADERS.about],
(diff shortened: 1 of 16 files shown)


simaob:
Hey Marta, forgot to ask before, what's the `content_type` field on the contents?

aabdaab:
It's for potential content manipulation on the frontend. Now, the contents are rendered in the order they are added on the admin panel. Although, if we'd like to have more control over what is rendered where, we could use `content_type` for that, it's supposed to act like a unique key on the FE. But it's not in place though and would require modifying the pages partial on-demand.

simaob:
Makes sense, we might want to add it as a dropdown and fixed set of values. But good for now!

tsubik:
I would also use friendly_id to create slugs from title. I don't think they will want to set that by hand (where we should make sure only allowed characters are being used)

tsubik:
I changed the stimulus part a little bit and added link as a title for admin page index.

So, the nested-list-controller would be more generic, to avoid clashes with inner nested-list-controllers there is a possibility to set `data-new-record-id` on the controller DOM element (https://github.com/Vizzuality/laws_and_pathways/pull/140/commits/b0d4382fbc39108602ae1b5a02169ec66d1f139b).

simaob:
We could leave a hint on the form for them to set slugs for new pages the way they want. If they are unique and we don't allow editing (for the base ones that we create ourselves), they could use whatever, no?

aabdaab:
I will rebase with develop now and look into the `friendly_id` ☝️

aabdaab:
Done ☝️ 
Removed the `slug` from Page form, modified the rake task accordingly and added a `friendly_id` for the Pages.

lib/tasks/static_pages.rake
```diff
@@ -4,21 +4,21 @@ namespace :static_pages do
...
+      ['FAQ', MENU_HEADERS.about],
+      ['Methodology', MENU_HEADERS.tpi_tool],
+      ['Data Background', MENU_HEADERS.tpi_tool]
     ]
 
     supporters_content = %w(partners supporters)
 
-    pages.each do |slug, title, menu_header|
+    pages.each do |title, menu_header|
       next if Page.find_by(slug: slug)
```
simaob: should this find be by title?

aabdaab: yeah, let me fix that quickly

aabdaab: just pushed a fix

PR #144, opened 2019-11-29
Feature/add tpi news

simaob:
This Pull Request is the first step to migrate News articles from the old TPI site to the new.

It adds the model the management page and the import file and script.

Need to check if we should add an image field to it or not, as some of their news had images. For now because the files are still available on the Wordpress the images are still linked to!


rapzo:
Is this for publications? If so, avatar/picture is missing.
We still need one for quotes/testimonials

simaob:
On the Publications and News we will show publications and News, these are the news. The fields in them are a little different, so I thought separate models made more sense, then we could "merge" them on the API. But let me know what you think?
Even if we go with a single model, we could merge this one so they could start reviewing the content for the news. And afterwards I'll import the publications as well.

Regarding testimonials, we'll need a separate model for that!

rapzo:
Gotchu. Let's head on then!

PR #146, opened 2019-11-29
Feature/tpi charts last reported year

tsubik:
- fixing seed of cp_unit for a sector
- adding unit to cp charts, sector one and company cp assessment one
- adding reported year distinction

image

I know that this sublegend (projected/targeted) should be below normal legend, but this was the quickest way and we will have to rewrite normal legend outside of highcharts anyway.

### DIFF
--- a/app/services/api/charts/sector.rb
+++ b/app/services/api/charts/sector.rb
@@ -141,10 +141,17 @@ module Api
       end
 
       def emissions_data_from_company(company)
+        assessment = company.latest_cp_assessment
+
         {
           name: company.name,
-          data: company.latest_cp_assessment&.emissions&.transform_keys(&:to_i),
-          lineWidth: 4
+          data: assessment&.emissions&.transform_keys(&:to_i),
+          zoneAxis: 'x',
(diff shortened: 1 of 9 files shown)


tsubik:
Make sure that sectors have units on staging/prod.

tsubik:
I also see there is something wrong with CP assessment date, and need to fix specs here.

app/services/api/charts/sector.rb
```diff
@@ -141,10 +141,17 @@ def emissions_data_from_sector_benchmarks
...
 
       def emissions_data_from_company(company)
+        assessment = company.latest_cp_assessment
+
         {
           name: company.name,
-          data: company.latest_cp_assessment&.emissions&.transform_keys(&:to_i),
-          lineWidth: 4
+          data: assessment&.emissions&.transform_keys(&:to_i),
+          zoneAxis: 'x',
```
kowal: I know it was that before, but we should separate data layer from presentation layer somehow.

tsubik:
no time for that, maybe later

PR #147, opened 2019-11-29
annotate models

kowal:


app/models/cp/assessment.rb
```diff
@@ -12,7 +12,6 @@
 #  updated_at         :datetime         not null
 #  discarded_at       :datetime
 #  last_reported_year :integer
-#  cp_alignment       :string
```
tsubik: shouldn;t be removed

kowal: ah, let me re-check this again

kowal: fixed ;)

PR #149, opened 2019-12-01
(feature): add user roles

kowal:
Groundwork before adding authorization rules:

* add `cancancan` gem
* add `role` to `AdminUser` model
* updates admin pages for users
* get rid of `update_attributes is deprecated` rails warning
* add missing specs

Pivotal: https://www.pivotaltracker.com/n/projects/2362592/stories/169121017

### DIFF
--- a/app/admin/admin_users.rb
+++ b/app/admin/admin_users.rb
@@ -13,7 +13,7 @@ ActiveAdmin.register AdminUser do
     column :gravatar
     column :email
     column :full_name
-    column :role
+    column(:role) { |user| user.role.titleize }
     column :created_at
     actions
   end
@@ -25,7 +25,7 @@ ActiveAdmin.register AdminUser do
       row :first_name
       row :last_name
       row :full_name
-      row :role
+      row(:role) { |user| user.role.titleize }
       row :gravatar
       row :updated_at
       row :created_at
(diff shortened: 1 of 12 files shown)


app/admin/admin_users.rb
```diff
@@ -72,7 +81,8 @@
   controller do
     # update without password if not provided
     def update_resource(object, attributes)
-      update_method = attributes.first[:password].present? ? :update_attributes : :update_without_password
+      update_method = attributes.first[:password].present? ? :update : :update_without_password
```
kowal: FYI: 'update_attributes' are deprecated in Rails 6 (will be removed in Rails 6.1) - https://github.com/rails/rails/pull/31998.

app/models/ability.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class Ability
+  include CanCan::Ability
+
+  def initialize(user)
+    @user = user
+
+    if @user
+      send(@user.role.to_sym)
```
kowal: this is not used yet, will be implemented in https://www.pivotaltracker.com/n/projects/2362592/stories/169121082

app/admin/admin_users.rb
tsubik: :+1:

simaob:
rebased and will merge now.

PR #151, opened 2019-12-02
Add publications model and management

simaob:


app/decorators/publication_decorator.rb
```diff
@@ -0,0 +1,7 @@
+class PublicationDecorator < Draper::Decorator
+  delegate_all
+
+  def title_link
```
tsubik: You can add file link here too. Take a look at DataUploadDecorator, on Publications show page there should be downloadable link for the file.

tsubik:
We gonna have to add more validation rules for things like thumbnails, like image minimum size. Also it could be processed by ImageMagick to a valid size after uploading. This is for other story though.

simaob:
Yeah, I'll add a task about validations, and added the suggested file_link.

tsubik:
:+1:

PR #154, opened 2019-12-02
Tpi/data fixes and other

tsubik:
- Updating CSVs for TPI data
- fixing import of cp assessment date
- various fixes for not loading charts
- company page - links for summary boxes to scroll down to respective section (mq assessment, or cp assessment)
- adding rake task to reimport tpi data `reimport:tpi` please check the task before running on prod, and see what it does, basically use with caution due to data loss.


tsubik:
I'm still working on changing CP alignment chart (CP performance - All sectors), but I can just create another PR with it as it's gonna take me one or 2 more hours to finish.

lib/tasks/reimport.rake
```diff
@@ -0,0 +1,11 @@
+namespace :reimport do
+  desc 'Reimport TPI data - USE WITH CAUTION'
+  task tpi: :environment do
+    Company.delete_all
```
kowal: maybe we can either:
- add required rake task param (like `rake reimport:tpi confirm=true)` in order to actually invoke it; without param it could print info what would be done
- or.. make it interactive, show `R u sure (y/n)` prompt - it is meant to be run manually anyway, right?

I can imagine someone will run it without reading description.

app/services/api/charts/sector.rb
```diff
@@ -145,7 +145,7 @@ def emissions_data_from_company(company)
 
         {
           name: company.name,
-          data: assessment&.emissions&.transform_keys(&:to_i),
+          data: (assessment&.emissions || {}).transform_keys(&:to_i),
```
kowal: couple of lines above we have similar line `data: benchmark.emissions.transform_keys(&:to_i)` - maybe unify it?

kowal: like
```ruby
def emissions_data_as_numbers(emissions)
  return {} unless emissions

  emissions.transform_keys(&:to_i)
end

data: emissions_data_as_numbers(assessment&.emissions)
data: emissions_data_as_numbers(benchmark&.emissions)
```

spec/services/api/charts/cp_assessment_spec.rb
```diff
@@ -68,7 +72,8 @@
           cp_benchmarks_data = subject.emissions_data.find { |s| s[:name] == 'scenario' }[:data]
 
           expect(company_sector_a_1_data).to eq(2017 => 90.0, 2018 => 90.0, 2019 => 110.0)
-          expect(sector_average_data).to eq(2017 => (90.0 + 40.0) / 2, 2018 => (90.0 + 50.0) / 2, 2019 => 110.0)
+          # sector average_data is only up to last_reported_year
```
kowal: :+1:

app/services/api/charts/sector.rb
tsubik: :+1:

lib/tasks/reimport.rake
tsubik: I think that's not necessary, access to production is limited.

app/services/api/presenters/company.rb
```diff
@@ -25,6 +25,10 @@ def summary
         )
       end
 
+      def model
```
kowal: Since we already have here `#cp_alignment` method, maybe also add smth like `#has_cp_alignment?` so it will be more readable in the template. What do you think?

tsubik:
@simaob TPI all sectors CP performance chart is updated, the only thing that is missing are those labels 
image

Will try to figure this out.

app/services/api/charts/cp_performance.rb
```diff
@@ -0,0 +1,43 @@
...
+      #       name: 'Paris Pledges',
+      #       data: [ ['Coal Mining', 65], ['Steel', 26] ]
+      #     }
+      #   ]
+      def cp_performance_all_sectors_data
+        all_companies = Company.published.includes(:sector, :latest_cp_assessment)
+
+        cp_alignment_data = {}
+        all_companies.each do |company|
+          next if company.cp_alignment.nil?
```
kowal: we could change `Company.published` to `Company.published.with_cp_aligment` and remove `next if ..` from here.

tsubik: Right

tsubik: hmm, actually how to build `with_cp_alignment` scope, this depends on latest_cp_assessment. Dunno, maybe I'm already too tired today.

app/services/seed/tpi_data.rb
```diff
@@ -0,0 +1,48 @@
...
+    include Singleton
+
+    class << self
+      delegate :call, to: :instance
+    end
+
+    # rubocop:disable Metrics/AbcSize
+    def call
+      # import NewsArticles
+      TimedLogger.log('Import news articles') do
```
simaob: I'm going to move the NewsArticles out of this main task. As they already started looking at them in prod, so wouldn't want to risk removing it.

simaob: Actually I'll just comment it for now. won't need to run them in prod.

tsubik: Should be actually moved out of this tpi_data class back to seeds.rb

simaob: yeah, we can probably get rid of it completely soon. Or maybe once we "lock up" the seeds to not run in prod. As we won't need to run this there, we can keep it for the development env later. I'll move it.

PR #155, opened 2019-12-03
Feature/bubble chart improvements

tsubik:
Small bubble charts improvements
- making more space, removing paddings, using width, height 100% of cell for svg element
- changing sizes of bubbles

image


tsubik:
Another PR should be about stopping that GForce animation at some point, and maybe even not showing it at all to be climate-friendly :D and not make page jitter.

PR #156, opened 2019-12-03
Feature/cclow search results

batraisk:


tsubik:
We should only show 10 entries and have this load more

image

Not sure how to implement this, but it could be done in another story.

aabdaab:
About Tomek's comment about lines truncation, if you want, take a look at the `publications.scss` file on `content__title` class. It's relying on the `max-height` per text (you assume how many lines of text you want to have visible and assuming that your `line-height` is 1rem for example, you multiple it by 3 if you want to have 3 lines of text visible) and then the `before` pseudoclass is the ellipsis you want to show and the `after` one is hiding that ellipsis in case the text is exactly 3 lines long or shorter.
Improvement of this would be showing `[...]` just next to the text so if you want to tackle that, go ahead 🙌

app/javascript/shared/controllers/content_list_controller.js
```diff
@@ -0,0 +1,38 @@
...
+      this.items[i].style.display = 'none';
+    }
+    $(this.buttonEl).click(this.loadMore.bind(this));
+  }
+
+  loadMore() {
+    this.showCount += this.step;
+
+    for (let i = 0; i < this.items.length; i += 1) {
+      this.items[i].style.display = this.showCount <= i ? 'none' : 'block';
```
tsubik: hmm, would be better to just load first 10 from the server and then load more, but let's have it in the next version.

PR #158, opened 2019-12-03
Hook publications

aabdaab:
This PR hooks up the publications and news articles in the FE.
* Display publications and news from the backend
* Fix styles for publications
* Add image && thumbnail to `NewsArticle` model
* Add show page for single publication/news
* Truncate the article title (`[...]` -> css way)

image


app/admin/news_article.rb
```diff
@@ -7,7 +7,7 @@
   decorate_with NewsArticleDecorator
 
   permit_params :title, :content, :publication_date, :article_type,
-                :created_by_id, :updated_by_id
+                :created_by_id, :updated_by_id, :thumbnail, :image
```
tsubik: Do we need different thumbnail and different image?

tsubik: Or this should be done by active storage image manipulation plugin

app/views/tpi/publications/_promoted.html.erb
```diff
@@ -1,70 +1,30 @@
...
-
-        <div class="content__title">
-          <a class="link is-strong">New research shows only two large oil & gas companies have long-term, low-carbon ambitions</a>
-        </div>
-
-        <div class="content__message">
-          <p>
-            Innovative new research assesses the carbon performance of the world’s ten largest public oil and gas companies, taking into […]
-          </p>
+  <div class="publications__grid">
```
rapzo: IMHO this will make it harder for mobile.

aabdaab: what is your concern? I just pushed the responsive version; it's only changing the `grid-template-columns` property

app/admin/news_article.rb
aabdaab: true, I'll fix that up

app/views/tpi/publications/_promoted.html.erb
rapzo: It was easier with bulma's flex grid. Don't see the point of overriding everything is all.

aabdaab: we had code for three columns only without wrapping up to the next rows; CSS grid handles that use case perfectly and it's very responsive. If we decide that instead of three columns, we want to have 4,5,6, you name it, you need to change one character; `grid-template-columns: repeat(3, 1fr);` 3 to 4,5,6, etc.
I don't mind using bulma, it's not for every use case though. Using grid here for much easier; otherwise with Bulma I'd probably have to change html structure.

rapzo: I understand that for desktop, but for mobile it's just one each row. Not so convenient. But i'd like to be proved wrong.

aabdaab: I thought I just did 😅
I don't understand what's inconvenient here. for having responsive implementation you don't need to change anything in the html structure. Everything you need to make it responsive for desktop is in the `@media tablet-portrait` block, as I tried to make it mobile-first. If we wanted to have even more breakpoints, desktop large, smaller mobile etc. we'd also need to change just this one CSS property. To me, it's very convenient - everything is in the CSS.

aabdaab: Take a look here: https://github.com/Vizzuality/laws_and_pathways/pull/158/files#diff-a314848eddcc2930479906ab6a4efa76R158
(linking it here as I added it after we started this conversation)

rapzo: Ah! Right! I've seen it!
Proved wrong.
P.S.: you can use an if/else clause with bulma's breakpoint variables instead of adding media queries

aabdaab: ah, true! will change that, thanks

simaob:
Thanks, looks good to me! =)

PR #163, opened 2019-12-04
Feature/add sectors and tags to pubs and news

simaob:
Adds tags and sectors to publications.
Fixes showing only 3 publications and news on the homepage.


tsubik:
It's from ActiveAdminAddons, but that's formtastic custom input, you can add any custom input you like.

PR #165, opened 2019-12-04
feature(tpi): add search bar

rapzo:
- click outside handled
- enter key handled
- escape key handled
- missing what to do on enter and with what


tsubik:
Do we need those icons `png` files? Any reason to keep that as we are using only svgs. Maybe we should move all icons to `icons` folder. Only svgs. Thoughts?

rapzo:
I fault there. I downloaded all the assets available in invisionapp and added them as i needed. Since i didn't want to be responsible for missing one version or the other, I've been shoving them all in.
Any cleaning sounds like a good idea.

simaob:
I'm wondering if for the actual search if we could just direct users to `/tpi/search?query=search-term` and then we load the data from the server into that page? Or is it better to handle in react?

rapzo:
That was my thought too, hence supplying the search url as a prop in order to redirect there. But since there's no search page, left it blank. Let me know if you want me to tackle that.

tsubik:
I think changing the location to search page should be good.

rapzo:
Is there any ujs turbolinks shenanigans that can help us there? Otherwise it will be a full page request.

tsubik:
We don't use Turbolinks as no-one has time to resolve issues with them :D

tsubik:
Full page redirect is fine for now.

simaob:
Lima can you setup the redirect, I can handle the search results page.

simaob:
on it!

rapzo:
There's an updated branch (not my fork), so feel free to push stuff there. Once you do it i'll look at what may be missing visually (although i believe those publications in the spec are useless for this matter, but look good... you choose).

simaob:
Search is hooked, Lima will give the results page some love!

PR #167, opened 2019-12-04
setup appsignal

tsubik:
I think it should be it, at least backend side. Gonna install on the frontend later, as I guess I need those environments created first, not sure how to do it in AppSignal UI, but after deploy both should be created.


simaob:
the spec is failing because of me I think. I fixed in the downloads fix PR.

tsubik:
ok

PR #168, opened 2019-12-04
Feature/improve downloads

simaob:
Following this thread: https://basecamp.com/1756858/projects/16557827/messages/88926209#comment_736046525

I'm going to check if we have the units on production as it should be showing up.

### DIFF
--- a/app/controllers/tpi/sectors_controller.rb
+++ b/app/controllers/tpi/sectors_controller.rb
@@ -49,7 +49,6 @@ module TPI
     end
 
     def user_download_all
-      timestamp = Time.now.strftime('%d%m%Y')
       companies_ids = Company.published.select(:id).where(sector_id: @sectors.pluck(:id))
       mq_assessments = MQ::Assessment
         .where(company_id: companies_ids)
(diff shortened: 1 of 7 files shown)


simaob:
yes, we didn't run the units import. Will add to the reimport task, was on the seeds.

tsubik:
Taking a look.

app/controllers/tpi/sectors_controller.rb
```diff
@@ -53,6 +53,8 @@ def user_download_all
       companies_ids = Company.published.select(:id).where(sector_id: @sectors.pluck(:id))
       mq_assessments = MQ::Assessment
         .where(company_id: companies_ids)
+        .joins(:company)
+        .order('companies.name ASC, assessment_date DESC')
```
tsubik: I think that should be done everywhere.

tsubik: we could sort it in ruby code, directly in exporters I think.

tsubik: or maybe better, with sql

tsubik:
I can take this over @simaob to change order everywhere and test more

simaob:
Thanks, do it!

simaob:
Is it worth doing further changes in a different PR, so that I can get this version into prod?

tsubik:
one sec

tsubik:
Done

tsubik:
Wait

tsubik:
Maybe CP benchamarks needs order too

tsubik:
one more sec

tsubik:
Done

PR #169, opened 2019-12-04
Fixing import files and adding warning

simaob:
Michal reported some data being imported wrongly. This should fix it. I've added some warnings to the import to compare IDs and Names, some names are not matching, but Michal is checking if it's a problem or just variations. The warning doesn't stop the import.


tsubik:
I see one is for a company I had to manually change, I might made a mistake (getting dizzy when looking at excel cells)

The other one is not, so maybe not my mistake ;]

tsubik:
So either name or id is wrong, we don't know.

simaob:
I've asked about it, let's see. I'll get this over to prod as it's better!

PR #170, opened 2019-12-04
Filtering publications

aabdaab:
This PR adds filtering for the publications and news page.

* Add react component `Filters`
* Handle query params on the controller
* Filtering by `OR` operator

For next PR/next task, tests are needed and refactor of the `handle...` function in the Filters component. There's a lot of duplication there but I tried to be as fast as possible. Also, I'd like to abstractify (is that even a word?) the filtering on the backend, without the if-ology.

Test in here:
`/tpi/publications`


app/controllers/tpi/publications_controller.rb
```diff
@@ -1,16 +1,61 @@
 module TPI
   class PublicationsController < TPIController
+    # rubocop:disable Metrics/AbcSize
     def index
-      publications = Publication.order(publication_date: :desc)
-      news = NewsArticle.order(publication_date: :desc)
+      @tags = (
```
tsubik: I would move that code into service object.

Also, without API we could do it differently:
- When request format is js. `format.js` just return html with filtered list and replace list html on the page. I guess that would work just fine.

tsubik: I can try to rewrite it a little bit

PR #173, opened 2019-12-05
news and publications filter refactor

tsubik:
Proposed changes


simaob:
I've fixed the specs to match the implementation.

PR #177, opened 2019-12-06
Adds CCLOWPage model, uses STI

simaob:
* moves Page from to TPI to TPIPage
* updates TPI frontend to use the right pages


app/admin/cclow_page.rb
```diff
@@ -0,0 +1,67 @@
+ActiveAdmin.register CCLOWPage do
```
tsubik: I think if this is the same, just menu is different and class name we could unified it in one file, the same as we do with Tags. Could be done later.

db/migrate/20191206082828_add_type_to_pages.rb
```diff
@@ -0,0 +1,10 @@
+class AddTypeToPages < ActiveRecord::Migration[6.0]
```
tsubik: Add Page class definition

```
Page = Class.new(ActiveRecord::Base)
```

as a good practive when migrating data and using AR models.

app/admin/cclow_page.rb
simaob: let's leave it for later

db/migrate/20191206082828_add_type_to_pages.rb
simaob: Thanks, will add!

aabdaab:
I get an error both on trying to create new TPIPage and CCLOWPage:
image

aabdaab:
The creation and edition works 👍
However, we might need to add the type to the pages like it was in TPI. I can see on the design that some of the static pages links are in the header:
image
some in the footer:
image
and some of them in both places (like `About`, `Methodology`, etc.). So we need to add a possibility to assign one or more types to the page. As per design, the only two possible types would be `header` and `footer`.

simaob:
You mean the menu?

aabdaab:
yeah

simaob:
I'll add it.

simaob:
Enabled, added three options: Header, Footer, Both. Feel free to change that to what makes most sense.

simaob:
In the `app/models/cclow_page.rb` file

PR #179, opened 2019-12-06
fix(tpi): fix active navbar highlight

rapzo:
- added a monstrosity, bug-filled to make it work
- requires testing for dropdown items


app/javascript/components/Navbar.js
```diff
@@ -137,16 +199,35 @@ const NavbarComponent = ({items, openSearchMode}) => {
...
+          title: PropTypes.string.isRequired,
+          slug: PropTypes.string.isRequired
+        })
+      ),
+      hasIcon: PropTypes.bool,
+      className: PropTypes.string
+    })
+  ).isRequired,
+  openSearchMode: PropTypes.func.isRequired,
+  active: PropTypes.shape({
```
tsubik: Is it possible to just do it looking only at url?

rapzo: Sure. The important is that they are mapped correctly at the items' entry level (i don't know the rails way™ to add the path there except hardcoded).

tsubik: Maybe active should in inside `items`

tsubik: https://github.com/Vizzuality/laws_and_pathways/blob/develop/app/views/layouts/cclow/geography.html.erb#L24

rapzo: If they have different urls, definitely.
What do you suggest on how to deal with them? Hardcoded? Should we put it in a helper/service or something?

tsubik: ahh, that would return link tag

tsubik: but there are other helpers https://github.com/comfy/active_link_to#helper-methods

tsubik: that could be added in layourt header.html.erb

rapzo: Yeah, due to how search behaves (overriding the nav) everything got out of rails context, so we need to somehow supply it via props (but if via path we don't need to supply the active one though).

tsubik: with those helpers is possible to just get is_active based on rails path.

tsubik:
You can rebase with develop to get rid of eslint errors in Filters component

app/javascript/components/Navbar.js
rapzo: I was looking at `url_to` and `path_to` helper methods that are used internally in `link_to`. I'll give them a try.

aabdaab:
Could you rebase with develop? There's a page crashing on `/cclow`, probably doesn't happen on develop 🤔

rapzo:
Will do. But require some caution since it's from friday or something.
Also, i'll try to work out a better way to present each item.

app/views/layouts/tpi/_header.html.erb
```diff
@@ -7,23 +7,25 @@
     },
     {
       entry: 'Publications and news',
-      path: '/tpi/publications',
+      path: url_for(controller: 'publications', only_path: true)
     },
     {
       entry: 'About',
       content: @about_pages,
+      path: url_for(controller: 'home', action: 'about', only_path: true)
```
tsubik: why not use `tpi_about_path`? the same for `tpi_publications_path`, `tpi_search_path`

tsubik:
Not sure what happened but I have two about pages which are displaying in mobile menu

image

But are not displaying in desktop navbar

image

Also those dynamic pages are not highlighted

app/views/layouts/tpi/_header.html.erb
```diff
@@ -7,23 +7,25 @@
...
     },
     {
       entry: 'Newsletter',
-      path: tpi_newsletter_path,
+      path: tpi_newsletter_path
```
tsubik: Instead of checking on frontend if path contains another, you can rewrite this to
```
{
      entry: 'Newsletter',
      path: tpi_newsletter_path,	  
      active: is_active_link?(tpi_newsletter_path)
}
```

tsubik: that will also work with nested rails routes with inclusive option (but we don't have those I guess)

tsubik: I also think that the same, checking if @tpi_tool_pages, and @about_pages, could be done with that `is_active_link?` helper too.

tsubik: I can help with that, but tomorrow my time will be very sparse, only in the evening (maybe)

rapzo: Wouldn't be this little helper be enough?
https://apidock.com/rails/ActionView/Helpers/UrlHelper/current_page%3F

tsubik: hmm, maybe this would be better :+1:

tsubik:
@rapzo What is the update on this?

rapzo:
Closing. Opening in a new PR.

PR #180, opened 2019-12-06
Starting by tidying up the Targets table

simaob:
* removed target_scopes table;
* added model to handle target's scopes like tags;
* small tweak in that form's order;


tsubik:
:eyes:

tsubik:
I haven't checked import/export though

tsubik:
Maybe upload test could be updated to check if scopes are correctly created.

tsubik:
Actually, nvm, could be done later. I think there is a task to refactor data uploads test. That would be anti-stress, leisure task, can pick it up next week.

simaob:
Yeah, that works. I'm trying to add the imports, but it's taking longer than I would've hoped. But it's getting there.

tsubik:
Yeah, half of import test are checking the data, but this one wasn't so no time to rewrite it now.

PR #182, opened 2019-12-06
Add tooltip for the CCLOW map

aabdaab:
* Tooltip appears on bubble click
* Set stroke for clicked circle
* Tooltip is clickable - set a link inside
* Style the tooltip
* Placeholder data

image

### DIFF
--- a/app/assets/stylesheets/cclow/_world-map.scss
+++ b/app/assets/stylesheets/cclow/_world-map.scss
@@ -1,3 +1,6 @@
+@import "colors";
+@import "variables";
+
 .world-map {
   &__container {
     position: relative;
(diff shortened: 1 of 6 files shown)


simaob:
Hey! =) Is it too confusing if it appears on hover? On the ToDo link you could point to the country page directly: https://laws-pathways-staging.vizzuality.com/cclow/geographies/143

aabdaab:
On the design it appears only on click: https://projects.invisionapp.com/d/main#/console/18784734/391067164/inspect

aabdaab:
And the placeholder stuff is there cause there's no data, so once it's hooked up it should be pointing to the country of the marker

app/assets/stylesheets/cclow/_world-map.scss
```diff
@@ -8,6 +11,58 @@
     background-color: $grey-light;
   }
 
+  &__tooltip {
+    background: $white !important;
+    color: $storm-gray !important;
```
aabdaab: both needed to be `!important` to override the 3rd party library

app/javascript/components/world-map/WorldMap.js
```diff
@@ -20,6 +21,26 @@ const PetersGall = geoCylindricalEqualArea().parallel(45);
...
+        <div className="world-map__tooltip-row">
+          <p className="world-map__tooltip-text">Greenhouse gas emissions (% of global emissions)</p>
+          <p className="world-map__tooltip-number">{emissions}</p>
+        </div>
+        <a href="https://todo-link-to-that-country" className="world-map__tooltip-link">{link}</a>
+      </div>
+    );
+  };
+
+  const [content, setTooltipContent] = useState('');
```
tsubik: I think I will move that into that reducer, we should either use reducer or state I guess?

PR #183, opened 2019-12-07
Feature/migrate laws data

simaob:
This PR creates a rake task to import the Laws Data, it's still not complete, but almost there, it brings:

* import legislations;
* import litigations;
* import targets;
* import events;

It still lacks:

* import litigation sides;
* import legislation files;
* import litigation files;

The sources files for legislation and litigations are already added to the repo, but still need to do the script. Later today or tomorrow, most likely.


simaob:
and I think I need to fix the dates import, because it's messed like it was on TPI.

tsubik:
Shouldn't we remove geography from Litigation and only leave Jurisdiction? :thinking:

tsubik:
Ahh, ok, I see some changes and jurisdiction is now just a text.

app/admin/legislations.rb
```diff
@@ -58,10 +56,9 @@
         attributes_table do
           row :title
           row :description
-          row :date_passed
           row :geography
           row 'Parent legislation', &:parent
-          row :sector
+          list_row 'Sectors', :laws_sector_links
```
tsubik: whatt? ;]

simaob: not needed? =D

simaob:
thanks for the fixes! =D

did you spot more things? Was going to look at the dates issue with events.

tsubik:
Yeah, still frontend is erroring in some places. We need to make geography, not optional now? International cases should be connected to international geography.

app/admin/legislations.rb
tsubik: Didn't know, that it could be in many sectors now.

tsubik: last time changes

simaob: yes... super fun :/ basically they didn't do the classification and told it could be found from the text by matching with a list then it leads to multiple sectors of course ...

simaob:
OK, I'll do a run through the app see crashes and add things as needed.

app/admin/legislations.rb
tsubik: :+1:

simaob:
I've fixed the specs and the date issues with the events. I would leave fixing the front end for a different PR. And before getting there I'll do the litigation sides in a different PR as well.
But if you are happy I can put this on staging for Michal to have a look

tsubik:
Yeah, you can do it.

PR #187, opened 2019-12-08
Feature/map data

tsubik:
Still WIP

List of tasks to be done here https://www.pivotaltracker.com/n/projects/2362592


tsubik:
I think it's ready, with the country page

image

simaob:
we has map! \o/

PR #191, opened 2019-12-09
Add correct logo assets to CCLOW header, add missing links

aabdaab:
* Add separate logos for LSE and Grantham
* Link logos to websites
* Add some missing links (contact form in the footer i.e)


simaob:
checking

PR #192, opened 2019-12-09
Add most recent additions in CCLOW search dropdown

aabdaab:
* Add most recent additions of laws, litigations and climate targets in the search dropdown
* Most recent = one month old
* Check against `created_at` attribute

### DIFF
--- a/app/controllers/concerns/cclow/search_controller.rb
+++ b/app/controllers/concerns/cclow/search_controller.rb
@@ -3,7 +3,7 @@ module CCLOW
     extend ActiveSupport::Concern
 
     included do
-      before_action :set_search_attributes
+      before_action :set_search_attributes, :set_recent_additions_attributes
     end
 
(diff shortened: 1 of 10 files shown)


simaob:
When I open the dropdown it shows me the total for both of those "recent additions", will it only update if I type a search?

aabdaab:
Because it's checking against the `created_at` and all of the records were created this month. Should we limit the number of the recent additions?

simaob:
ah true, let's use something different, see my sorting on `app/services/api/latest_additions.rb`

simaob:
Basing it on the events: https://github.com/Vizzuality/laws_and_pathways/blob/develop/app/services/api/latest_additions.rb#L17

aabdaab:
Changed the recently added legislations and litigations to use the `LatestAdditions` service but in the future we should probably change it to pick up from `passed` and `started` scopes (legislation and litigation respectively)

PR #194, opened 2019-12-09
Feature/cclow search filter

batraisk:


simaob:
nice to have some filters! =) good proof of concept to start with :D

PR #197, opened 2019-12-09
Adds targets API route per country

simaob:
This PR creates a JSON endpoint to return the Targets for a given country.

Usage:
http://localhost:3000/cclow/api/targets/:iso

It copies the same formatting as the old one: Example here: https://www.climatewatchdata.org/api/v1/lse_laws_and_policies/BRA

I'm a bit tired and confused, so might have written the most ugly code ever. I had to pass the request protocol and host to get the right path, but maybe I could have added some helper to the model to access it directly. Thoughts? Thanks!

### DIFF
--- a/app/models/target.rb
+++ b/app/models/target.rb
@@ -71,24 +71,4 @@ class Target < ApplicationRecord
     parts = [geography.name, target_type&.humanize, year]
     parts.compact.join(' - ')
   end
-
-  def as_json(options)
-    path = options[:path]
-    {
-      id: id,
-      iso_code3: geography.iso,
-      doc_type: source,
-      type: target_type,
-      scope: scopes.map(&:name).join(', '),
-      sector: sector&.name,
-      description: description,
-      sources: legislations.map do |l|
-        {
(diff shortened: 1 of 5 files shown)


app/models/target.rb
```diff
@@ -71,4 +71,24 @@ def to_s
...
+      doc_type: source,
+      type: target_type,
+      scope: scopes.map(&:name).join(', '),
+      sector: sector&.name,
+      description: description,
+      sources: legislations.map do |l|
+        {
+          id: l.id,
+          title: l.title,
+          link: "#{path}/cclow/geographies/#{geography.id}/#{l.route}/#{id}"
```
tsubik: Rails.application.routes.url_helpers could help with taht

tsubik:
I think I would move that code from model to controller. As this json formar is more specific to only that controller

app/models/target.rb
simaob: gonna move this to the controller.

simaob:
updated and added specs as a bonus!

PR #198, opened 2019-12-09
Cclow/search to the backend pgsearch

tsubik:
After experimenting with custom sql query using full text search, I decided to use `pg_search` gem, and it works nicely. I need to experiment with more fuzzy like options that this gem provides. The results will be obiously different than before, that's my only concern.

I'm also refactoring cclow controllers here, fixing issues with searching using query + regions + any other filter we will add later.

The other option for that slow loading search could be still use fuzzy on frontend, but load all data from API, not on page load. That should work just fine now, but maybe won't be that scalable, and also for results on other pages it is better to just use single `q` param and not list of all `ids`


tsubik:
@simaob I'm not sure what is the recipe for most recent addition to climate targets. Litigations and Legislation I've done based on the current code here https://github.com/Vizzuality/laws_and_pathways/pull/198/commits/e1086619db96ee0528dfb23a15cc82a9a71cb03b

simaob:
Yeah, I think we need to discuss this in more detail with Michal. If they will add events to all the laws and litigations, then this is perfect. I think that's what makes sense anyway.

tsubik:
image

Breadcrumbs and page title with filter presented fixed

tsubik:
I think this is ready for review although I need to:
- optimize full text
- try with some fuzzy search not full text
- some refactor is needed, all results react components could be combined I guess

tsubik:
Also some "Loading..." for filters would be good, or at least there is some state when filters are hidden, don't have time now to debug this.

tsubik:
Tests would be nice too :/

simaob:
Pretty cool! =D I need to test a bit further, but it's looking great =)

tsubik:
I also see there is highlight option in full text search, so there is room for improvement, to not show description but that highlighted fragment where searched text was found.

tsubik:
If I find a little time today  I will add indexes, and check other pg_search mode, because it's not a fuzzy one now.

tsubik:
Actually, I won't have time today, soo much stuff to pack for tomorrow :fearful:

aabdaab:
Cool 🙌 ! I'm curious about the stats in AppSignal after we merge it!
There is some lag though when typing the search key; for a moment it shows results for the previous search key:
screencast 2019-12-12 16-37-31

tsubik:
Yeah, I see that too.

PR #199, opened 2019-12-10
fix(cclw): add lil world map picture to hero unit

rapzo:


aabdaab:
hey 👋 I saw that you added the background image to the assets but haven't changed that in the code. Can we tackle this in that PR as well? Also, the map image looks of low quality on the website compared to what we have in the design, can you double-check the assets?
See below: (the left one is from the website, on the right - design)
image

rapzo:
I belive it should be fixed now. Filenames were messed up.

PR #202, opened 2019-12-12
Feature/cclow search filter

batraisk:


simaob:
Are the filters hooked with the backend? They don't seem to be. Selecting a country doesn't filter the list of laws.

tsubik:
I think that's my bad, I saw a region before, missed the country. So, that's filtering by region only.

tsubik:
Taggin wasn't cleaned up before import, crap :/

image

tsubik:
or the delimiter should be ';' or maybe there were 2 delimiters :P

simaob:
Yes there's some crap in there. But most of mine look good. I can sort it out in the prod console later =D split and save. =)

tsubik:
When filtering by Tags there should be `AND` not `OR` right?

tsubik:
or maybe OR

simaob:
that's always a tough question... @faustoperez what do you think? Tags are "has this AND that" or "has this OR that"?

tsubik:
I see I must removed filter by geography.

tsubik:
Countries should be `OR` I guess, so maybe tags too for consitency.

tsubik:
I'm gonna bring back the `geography` filter @batraisk

simaob:
Agreed, good thinking. OR it is.

tsubik:
Done, geography should work.

simaob:
Any other feedback? It looks pretty good to me! =)

simaob:
Yes, sounds good! Can come in a separate PR!

tsubik:
Let's how it works on staging.

batraisk:
thank you @tsubik :)

PR #203, opened 2019-12-14
authorization rules

kowal:
- [x] define abilities
- [x] update visibility status sidebar
- [x] update index pages to conditionally display links
- [x] add ability specs
- [x] fix issue with abilities for decorated resources - dc1bfaa
- [x] batch actions - do not display menu items which aren't allowed - 61c2145
- [x] batch actions - fix issue with empty `batch_action_collection` - replaced with `scoped_collection`
- [x] managing users - publisher can manage only roles from their area - e42c43f 
- [x] data-upload - editor users can't 'publish' resources via import - cafba21

### DIFF
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -174,7 +174,7 @@ ActiveAdmin.register Company do
         column { f.input :geography }
         column { f.input :headquarters_geography }
         column do
-          f.input :visibility_status, as: :select
+          f.input :visibility_status, visibility_status_select_options
         end
       end
 
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -7,7 +7,7 @@ ActiveAdmin.register Company do
   config.batch_actions = false
 
   publishable_scopes
-  publishable_sidebar only: :show
+  publishable_resource_sidebar
 
   permit_params :name, :isin, :sector_id, :geography_id, :headquarters_geography_id,
                 :ca100, :market_cap_group, :visibility_status, :sedol,
--- a/app/admin/litigations.rb
+++ b/app/admin/litigations.rb
@@ -4,7 +4,7 @@ ActiveAdmin.register Litigation do
   decorate_with LitigationDecorator
 
   publishable_scopes
-  publishable_sidebar only: :show
+  publishable_resource_sidebar
 
   permit_params :title, :geography_id, :document_type, :jurisdiction,
                 :visibility_status, :summary, :at_issue,
--- a/app/admin/litigations.rb
+++ b/app/admin/litigations.rb
@@ -118,12 +118,14 @@ ActiveAdmin.register Litigation do
 
   form partial: 'form'
 
-  batch_action :publish,
-               priority: 1,
-               if: proc { current_scope&.name != 'Published' } do |ids|
-    publish_command = ::Command::Batch::Publish.new(batch_action_collection, ids)
+  can_batch_publish = proc { can?(:publish, Litigation) && current_scope&.name != 'Published' }
+  can_batch_archive = proc { can?(:archive, Litigation) && current_scope&.name != 'Archived' }
+  can_batch_destroy = proc { can?(:destroy, Litigation) }
 
-    message = if publish_command.call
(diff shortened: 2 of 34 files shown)


app/admin/companies.rb
```diff
@@ -7,7 +7,7 @@
   config.batch_actions = false
 
   publishable_scopes
-  publishable_sidebar only: :show
+  publishable_resource_sidebar
```
kowal: We display this sidebar on `show` pages only, it doesn't make sense for collection (we have bulk action for that).

app/admin/news_article.rb
```diff
@@ -82,9 +82,5 @@ def scoped_collection
     def apply_filtering(chain)
       super(chain).distinct
     end
-
-    def csv_format?
```
kowal: this wasn't used

app/helpers/select_helper.rb
```diff
@@ -9,6 +9,22 @@ def array_to_select_collection(array, transform_func = :humanize)
     end
   end
 
+  def user_roles_select_collection
+    array_to_select_collection(current_user_roles_to_select, :titleize)
+  end
+
+  # returns roles which current user can set when creating/editing other users
+  def current_user_roles_to_select
```
simaob: When editing a user profile and the current_user has one of the editor roles the dropdown is showing "Yes" and "No" as the options which are not valid options. I guess we can make the role not editable in this case. This will only happen if they are editing their own account.

kowal: Do we want this (disable role input when editing self) even for super admins?

kowal: fixed in 25e7562 - when editing self, role input is disabled

tsubik:
:eyes:

app/helpers/select_helper.rb
tsubik: Meaning that only other super user can edit someone's role? We can call that a security feature :)

app/helpers/select_helper.rb
```diff
@@ -9,6 +17,23 @@ def array_to_select_collection(array, transform_func = :humanize)
     end
   end
 
+  # Returns roles which current user can set when creating/editing other users
```
tsubik: I would personally allow only super users to create accounts, and edit roles.

tsubik: Just to make things easier

tsubik: Ohh, I see that's was on the story and I guess decided with the client, so nevermind.

tsubik:
I added one example user for each role for development environment.

tsubik:
Not sure what is a good way to have this sort of authorization on the attribute level and the current value of the attribute.

kowal:
> As an editor I was able to publish/unpublished entity by just changing the status value in edit form.

We can make **Visibility Status** select as read-only if current user is 'editor'.

>  I think the same could be done by using bulk import.

Need to think about this ..

app/admin/litigations.rb
```diff
@@ -118,12 +118,14 @@
 
   form partial: 'form'
 
-  batch_action :publish,
-               priority: 1,
-               if: proc { current_scope&.name != 'Published' } do |ids|
-    publish_command = ::Command::Batch::Publish.new(batch_action_collection, ids)
+  can_batch_publish = proc { can?(:publish, Litigation) && current_scope&.name != 'Published' }
```
kowal: we could probably extract those helpers somewhere

app/services/csv_import/geographies.rb
```diff
@@ -4,6 +4,8 @@ class Geographies < BaseImporter
 
     def import
       import_each_csv_row(csv) do |row|
+        check_permissions_for_row(row)
```
kowal: maybe we could move it into `check_permissions_for_row`, wdyt?

tsubik:
Simao had an idea of adding callback to the model. I could try to add that one. In `before_save` to check if `is_publishing?` if the current user can publish, if not then raise exception. No need to add special code to importers then.

simaob:
With your changes merged into this, are we good to go with the authorization, Tomasz?

tsubik:
I think so. We will see :)

PR #204, opened 2019-12-16
Responsive map

aabdaab:
* Add responsive styles for the Map
* Fix map controls styles
* Fix map dropdowns styles

image

Screenshot 2019-12-16 at 11 48 47


app/javascript/components/world-map/MapLegend.js
```diff
@@ -51,8 +51,8 @@ function MapLegend({ content, context, scales }) {
               <div>
                 <div className="rectangle" style={{backgroundColor: color}} />&nbsp;
 
-              {(i === 0) && `<${displayContextValue(color, 1)}`}
-              {(i === colors.length - 1) && `>${displayContextValue(color, 0)}`}
+                {(i === 0) && `<${displayContextValue(color, 1)}`}
```
tsubik: emacs keeps messing this up lol

app/javascript/components/world-map/WorldMap.js
```diff
@@ -58,6 +59,42 @@ function WorldMap({ zoomToGeographyIso }) {
   // react-select
   const setContentId = selectedOption => dispatch({ type: 'setContentId', payload: selectedOption.value });
   const setContextId = selectedOption => dispatch({ type: 'setContextId', payload: selectedOption.value });
+  const customStyles = {
```
tsubik: I'm totally not a fan of CSS in javascript, but as react-select went this way :man_shrugging:

tsubik: That's why I haven't styled it before, procrastinating things I don't like, but tell no-one :P

aabdaab:
As a bonus, I added here the fix for broken assets in the search results page. I'm not happy about that solution because we might end up adding more assets to the public folder if we want to use them a similar way but have failed to fix that with rails webpacker :/

PR #207, opened 2019-12-17
Feature/allow multiple sectors for publications

simaob:
Michal requested that we change the publications and news articles so that they support multiple sectors. In this PR I added this, updated backend, updated frontend, and specs.

These two models are very very similar, we probably should merge them into one. But not on this PR, sorry.


aabdaab:
hey! I've found one error due to the changes when going to the Publications index page and trying to filter by sectors:

image

simaob:
Think it's fixed now.

aabdaab:
👀

PR #215, opened 2019-12-21
Feature/cclow climate targets page

batraisk:


app/controllers/cclow/geography/climate_targets_controller.rb
```diff
@@ -5,20 +5,46 @@ class ClimateTargetsController < CCLOWController
...
+        @ndc_targets = CCLOW::TargetDecorator.decorate_collection(@ndc_targets)
+
+        @climate_targets = @geography.targets.published.where(sector: @sector).where.not(source: 'ndc')
+        @climate_targets = CCLOW::TargetDecorator.decorate_collection(@climate_targets)
+        render 'cclow/geography/climate_targets/target_sector'
+      end
+
+      private
+
+      def targets_list_by_sector
```
simaob: I think this should be a method inside of Geography, as it's quite specific to a single geography. And instead of using ruby we could query the results directly with something like:

`LawsSector.joins(:targets).where(targets: { geography_id: @geography}).select('laws_sectors.name, laws_sectors.id, targets.source, COUNT(*) AS count').group('laws_sectors.id', 'laws_sectors.name', 'targets.source')`, that would return the sector names and the count and then we can just iterate over it. What do you think?

app/views/cclow/geography/climate_targets/index.html.erb
```diff
@@ -1,87 +1,59 @@
...
-        <div class="nav">
-          <h5>NDC content</h5>
-          <span class="counter"><%= pluralize(@ndc_targets.count, 'target') %></span>
+    </li>
+    <div class="sectors-list">
+      <div class="columns">
+        <div class="column is-one-thirds"></div>
+        <div class="column is-one-thirds column-name">Targets in NDC content</div>
+        <div class="column is-one-thirds column-name">
+          <span>Targets in national</span><span>laws and policies</span>
```
simaob: why the two separate spans here? to break the lines?

app/controllers/cclow/geography/climate_targets_controller.rb
```diff
@@ -5,20 +5,46 @@ class ClimateTargetsController < CCLOWController
...
+        counts = Target.joins(:sector).where(geography: @geography).group('laws_sectors.name', :source).count
+        list = {}
+        counts.each do |key, value|
+          sector_name = key[0]
+          source_name = key[1]
+          next unless %w(law policy ndc).include?(source_name)
```
simaob: we could also add a where clause to exclude items that don't belong to one of these, right?

app/controllers/cclow/geography/climate_targets_controller.rb
simaob: hmmm, I think that tentative query would still need some work, but would be achievable =)

app/views/cclow/geography/climate_targets/index.html.erb
```diff
@@ -1,87 +1,59 @@
...
-      </div>
-    </li>
+            <% else %>
+              <div class="empty-circle"></div>
+            <% end %>
+          </div>
+        </a>
+      <% end %>
+
+      <% (LawsSector.select(:name).map(&:name) - @list.keys).each do |sector| %>
```
simaob: we should sort this alphabetically and maybe even just pass the keys to the query, instead of getting everything and then removing the ones not needed.

app/controllers/cclow/geography/climate_targets_controller.rb
simaob: we also need to ensure that we are counting the NDC targets from EU for the EU countries.

aabdaab:
ready again ☝️

PR #223, opened 2020-01-09
Add dotted timeline to a single point event

aabdaab:
* If there's only one event on the timeline, keep the dotted line
* All left-aligned as per client's request except when there is only one event in order to show the dotted timeline

**Single event**:

image

**Two events**: 
image

**More events**
image

### DIFF
--- a/app/assets/stylesheets/cclow/_events_timeline.scss
+++ b/app/assets/stylesheets/cclow/_events_timeline.scss
@@ -6,7 +6,6 @@ $arrow-btn-width: 48px;
   display: flex;
   position: relative;
   overflow: hidden;
-  margin-left: $arrow-btn-width;
   width: calc(100% - #{$arrow-btn-width * 2});
 
   .timeline {
--- a/app/assets/stylesheets/cclow/_events_timeline.scss
+++ b/app/assets/stylesheets/cclow/_events_timeline.scss
@@ -17,37 +17,37 @@ $arrow-btn-width: 48px;
     position: relative;
     text-align: center;
     width: 275px;
-  }
 
-  .time-point.time-point-multiple-events {
-    &:first-child {
-      text-align: left;
-      width: 145px;
-    }
-  }
-
-  .time-point-multiple-events {
-    &:last-child {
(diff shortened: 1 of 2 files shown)


app/javascript/components/EventsTimeline.js
```diff
@@ -93,7 +94,7 @@ const EventsTimeline = ({ events, options, isFiltered = false }) => {
         <div ref={eventsContainerEl} onScroll={checkButtons} className="events-container">
           <div className="timeline">
             {currentEvents.map((event) => (
-              <div key={event.title} className="time-point">
+              <div key={event.title} className={cx('time-point', { 'time-point-multiple-events': currentEvents.length > 1 })}>
```
tsubik: maybe with BEM modifier `time-point--multiple-events` I know we are not very consistent with using BEM ;)

but after that CSS could be written like
```
.time-pont {
   ....
   &--multiple-events {
   ...
   }
}
```

aabdaab: sure, yeah 👍

tsubik: actually maybe always could be written like `&-multiple-events` haven't checked that.

aabdaab:
added a commit with nesting ☝️

app/assets/stylesheets/cclow/_events_timeline.scss
```diff
@@ -17,37 +17,37 @@ $arrow-btn-width: 48px;
...
 
-  .time-point-multiple-events {
-    &:last-child {
-      .point {
-        &:after {
-          border: none;
-        }
+    &.time-point-multiple-events {
+      &:first-child {
+        text-align: left;
```
tsubik: this could be moved into first-child of &-multiple-events right?

tsubik: As I can see there is no need to check if both classes are on the element time-point and time-point-multiple-events, hmm

aabdaab:
fixed, gonna merge this

PR #224, opened 2020-01-09
fix exporting litigations

tsubik:
Quick fix for admin litigations export.

Geography is optional I see. Should it be? International cases do not have a geography. I thought there should be a special "International" geography, with flag and without region. That could be case for a new story to create it and properly apply filters when searching litigations.

### DIFF
--- a/app/admin/litigations.rb
+++ b/app/admin/litigations.rb
@@ -61,8 +61,8 @@ ActiveAdmin.register Litigation do
     column :id
     column :title
     column :document_type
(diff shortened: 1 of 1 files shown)


simaob:
Yes, you're right, Geography should be mandatory. And we might need to create the international geography, I thought we had it. 🤦‍♂

simaob:
actually we have it already: https://climate-laws.org/admin/geographies?q%5Bname_contains%5D=Inter&commit=Filter&order=name_asc we might need to link them.

PR #225, opened 2020-01-09
Fix navbar's active item highlight

rapzo:


rapzo:
Not my best ruby-fu applied here. Feel free to throw your tips on how to ruby ;-)

tsubik:
Thanks. I will take a look.

tsubik:
Just wondering about this "about page" :smile: 

There is `about` action in home controller, but about page is a dynamic one. I think it's kinda misleading. I will try to remove that empty page and see if the menu works after that.

tsubik:
Actually on prod I see there is no about page. That's just the menu header name.

rapzo:
Yes, it was my doings (the about action in the home controller and all)...
But, even if it doesn't exist, i thing the helper will always return false for its path analysis and go about the paths from the dynamic generated. It should be safe either to keep or remove it IMO.

app/helpers/application_helper.rb
```diff
@@ -16,4 +16,22 @@ def render_breadcrumb
...
   end
+
+  def active_menu_page?(**args)
+    unless args.nil?
+      has_pages = false
+      has_path = false
+
+      has_pages = args[:pages].any? { |page| current_page?(page[:slug]) } unless args[:pages].nil?
+
+      has_path = current_page?(args[:path]) unless args[:path].nil?
```
aabdaab: It's not a blocker but a matter of convention comment; `hasSomething` convention in javascript becomes `something?` in ruby - `?` in ruby indicates that a method returns a boolean value (like internal blank?, valid? etc).

tsubik: I'm actually refactoring. So all that will change.

aabdaab: 👍

tsubik: Trying to make sectors and companies controller actions to also highlight the TPI menu

rapzo: Oh... Didn't go over those... Sorry.

app/javascript/components/DropdownSelector.js
```diff
@@ -171,40 +189,62 @@ const DropdownSelector = ({ sectors, companies, selectedOption, defaultFilter =
             <div className="dropdown-selector__options-wrapper">
               {isFilterBySector && (
                 <div className="dropdown-selector__options">
-                  {options.map(option => (
-                    <div onClick={() => handleOptionClick(option)} className="dropdown-selector__option">
+                  {(options.length > 0 && options.map((option, i) => (
```
aabdaab: no need to check if it's `> 0` - `options.length` will evaluate to `false` if it's null or 0

app/javascript/components/DropdownSelector.js
```diff
@@ -171,40 +189,62 @@ const DropdownSelector = ({ sectors, companies, selectedOption, defaultFilter =
...
+                    <div
+                      onClick={() => handleOptionClick(option)}
+                      className="dropdown-selector__option"
+                      key={`${option.name}-${i}`}
+                    >
                       {option.name}
                     </div>
+                  ))) || (searchValue.length > 0 && options.length === 0 && (
```
aabdaab: same here: you can remove `> 0` bit and instead of explicitly checking if `options.length === 0` turn it into `!options.length`

app/javascript/components/DropdownSelector.js
```diff
@@ -171,40 +189,62 @@ const DropdownSelector = ({ sectors, companies, selectedOption, defaultFilter =
...
                   ))}
-                  {searchValue && !options.length && <div>No results found.</div>}
                 </div>
               )}
               {isFilterByCompany && (
                 <div className="dropdown-selector__options">
-                  {Object.keys(companiesBySector).map(sector => (
-                    <>
-                      <div className="dropdown-selector__sector-name">
+                  {(options.length > 0 && Object.keys(companiesBySector).map((sector, i) => (
```
aabdaab: and here

app/javascript/components/DropdownSelector.js
```diff
@@ -171,40 +189,62 @@ const DropdownSelector = ({ sectors, companies, selectedOption, defaultFilter =
...
+                          key={`${option.name}-${i}-${j}`}
                         >
                           {option.name}
                         </div>
                       ))}
-                    </>
-                  ))}
-                  {searchValue && !options.length && <div>No results found.</div>}
+                    </Fragment>
+                  ))) || (searchValue.length > 0 && options.length === 0 && (
```
aabdaab: I see that this is repeating, maybe move it to a variable

tsubik:
Ok, done some little ruby refactoring, hopefully didn't break anything.

app/javascript/components/DropdownSelector.js
rapzo: i know what you mean but readability speaking I only want something to happen when that condition is true. If options was an object with a map function and a length property, say, a string, the sole condition of checking the property would pass whereas the condition for checking its length won't.
TL;DR it's like `i++` vs `i += 1` readability vs habits

app/javascript/components/DropdownSelector.js
rapzo: Same motif as in the comment before.

rapzo:
Updated. Tested it and all seems good to go!

PR #226, opened 2020-01-09
Fix litigation-sides importer - overriding records issue

aabdaab:
Fix missing litigation sides issue: in [this case](https://climate-laws.org/cclow/geographies/9/litigation_cases/7092) we're missing `Side A`.
The issue was the importer which was overriding the first csv row with the second one based on the same litigation_id, connected_entity_id, connected_entity_type, party_type attributes (litigation_id and party_type here are the same and connected_entity_id && connected_entity_type are both null)
```
7092,A,European Commission,government
7092,B,Council for the European Union,government
```
We're gonna need to run the `rake reimport:cclow` task to account for these missing rows.
Test here: http://localhost:3000/cclow/geographies/9/litigation_cases/7092

### DIFF
--- a/app/services/csv_import/litigation_sides.rb
+++ b/app/services/csv_import/litigation_sides.rb
@@ -23,13 +23,7 @@ module CSVImport
     end
 
     def prepare_litigation_side(row)
-      find_record_by(:id, row) ||
-        LitigationSide.new(
-          litigation_id: row[:litigation_id],
-          connected_entity_id: row[:connected_entity_id],
-          connected_entity_type: row[:connected_entity_type],
-          party_type: row[:party_type]&.downcase
-        )
+      find_record_by(:id, row) || resource_klass.new
     end
 
     def litigation_side_attributes(row)
--- a/app/services/csv_import/litigation_sides.rb
+++ b/app/services/csv_import/litigation_sides.rb
@@ -24,7 +24,7 @@ module CSVImport
 
     def prepare_litigation_side(row)
       find_record_by(:id, row) ||
-        LitigationSide.find_or_initialize_by(
+        LitigationSide.new(
           litigation_id: row[:litigation_id],
           connected_entity_id: row[:connected_entity_id],
           connected_entity_type: row[:connected_entity_type],
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -374,23 +374,6 @@ ActiveRecord::Schema.define(version: 2020_01_10_103445) do
     t.index ["updated_by_id"], name: "index_litigations_on_updated_by_id"
   end
 
-  create_table "locations", force: :cascade do |t|
-    t.string "location_type", null: false
-    t.string "iso", null: false
-    t.string "name", null: false
-    t.string "slug", null: false
-    t.string "region", null: false
-    t.boolean "federal", default: false, null: false
-    t.text "federal_details"
-    t.text "approach_to_climate_change"
-    t.text "legislative_process"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-    t.index ["iso"], name: "index_locations_on_iso", unique: true
-    t.index ["region"], name: "index_locations_on_region"
-    t.index ["slug"], name: "index_locations_on_slug", unique: true
-  end
-
   create_table "mq_assessments", force: :cascade do |t|
     t.bigint "company_id"
     t.string "level", null: false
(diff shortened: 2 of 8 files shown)


aabdaab:
I see that the test is failing so it was done on purpose, what context am I missing here? We could also keep the `find_or_initialize_by` method but add a `side_type` to the list of attributes that the LitigationSide is being checked against

simaob:
Good catch! We should confirm with them if they made changes to the sides. And maybe it's best to create a separate reimport task to just import the sides. And avoid messing with the whole dataset.

aabdaab:
Yeah, a dedicated rake task for this is a good idea 👍

simaob:
I spoke too soon I guess. =D had just read not reviewed thoroughly.
Yes adding the side_type might do the trick. We were finding or initializing so that it can be updated with the importer.

simaob:
Let me have another think and look at the code again. The idea is to find the right rows to edit in case it exists to be updated. But need to confirm for the litigationSides if this makes sense. I'll go check it now. But do add the dedicated rake task!

aabdaab:
ok! thanks

simaob:
Looking at the data, I think it's worth adding the side_type to that find_or_initialize. It will ensure that the importer will add more rows, and not skip others. Which is a bug.
On the other hand there's not a lot of things to edit on the Litigation sides, if we pass the party_type, the connected_entity and the side_type, the only thing left to edit is the name. Not sure if this will be used a lot. Maybe it would be simpler to just add a new one each time, specially as there can be multiple side as, bs and cs. What do you think @tsubik ?

aabdaab:
now that I added a separate rake task I see that the `find_or_initialize_by` method makes sense cause without it, each time we run an import we duplicate the data. But the way it works in the global reimport is first we delete all the records so the find_by doesn't really make sense. So the question remains as now we have a separate importer which isn't deleting the rows first -- or should it? Then we wouldn't need find_or_initialize_by

simaob:
Our migration importers make use of these, but these are also used in the backend, so we want to ensure that they can add new data but also update existing data, hence the find_or_initialize_by. For a custom migration importer we could not use it. But check my previous comment. For this particular piece of data (LitigationSide) there won't be a lot of things to edit, I think, and because there can be multiple, it might make more sense to not have the edit version.

aabdaab:
It makes sense; I added a commit with the separate rake task `rake reimport:cclow_litigation_sides` and removed that failing test as it's not relevant anymore. Since we don't need editing, the rake task first deletes all litigation sides and then imports the whole litigation sides file again, so duplication won't occur.

tsubik:
Ohh crap. The find_or_initialize_by was more to prevent accidental duplications when someone will re-upload the same file once again through admin panel. For simple updates, there is `id` column. So as you have said Simao, we can either add missing attributes to find_or_initialize_by or just remove that completely from there. I have no strong preference, maybe we can remove it, will be more performant. Or use unique index on all columns to prevent duplicates :P, not sure what's the cost of this.

tsubik:
Or we can use combination of `litigation_id`, `side_type` and `name` in find_or_initialize_by, or maybe even sole `name` for a side should be unique in the context of `Litigation.`

app/services/csv_import/litigation_sides.rb
```diff
@@ -24,7 +24,7 @@ def resource_klass
 
     def prepare_litigation_side(row)
       find_record_by(:id, row) ||
-        LitigationSide.find_or_initialize_by(
+        LitigationSide.new(
```
tsubik: If not find_or_initalize then `resource_klass.new` because assign attributes from `row` is already done later.

tsubik: But as I wrote, maybe combination of `litigation_id`, `name`, `side_type` would be good to prevent accidental duplicates.

tsubik: Or actually unique index on those values could be more beneficial. Thoughts?

simaob: with a unique index it will stop duplicate rows from being added right? I think that's a good solution. So we assume that editing wouldn't happen in bulk like this but just additions, which I think is fair to say. Even to edit names of the sides, it probably makes more sense through the interface. Right?

aabdaab: I think so. Editing is tricky as it's hard to determine what should always stay the same in the csv file.

db/schema.rb
```diff
@@ -373,6 +374,23 @@
     t.index ["updated_by_id"], name: "index_litigations_on_updated_by_id"
   end
 
+  create_table "locations", force: :cascade do |t|
```
tsubik: Hey @aabdaab what happened here? Worth to reset the database. I'm doing that all the time. That's why the need I stress of the need of good and complete seeds :P

aabdaab: hmm, good question 🤔I ran migrations, let me reset

aabdaab: I resetted the DB but same result; I guess develop is missing this schema bits

tsubik: Locations table is old, now it's geography.

tsubik: You have to reset table when beigin on develop, then go back to your branch, and run migrations. If you reset it on your branch that's gonna create those locations and sectors tables.

app/services/csv_import/litigation_sides.rb
tsubik: You still can edit it through bulk feature, after exporting existing, those will have litigation side `id` attached and the corresponding record will be found.

tsubik:
You didn't have to remove the index, could be there to ensure no duplicates are there.

aabdaab:
Hmm, interesting that you're pointing that out. I had a dilemma about it and didn't want to repeat it on the db level as well. I just read [this article](https://thoughtbot.com/blog/validation-database-constraint-or-both) though and maybe we should it keep after all.

aabdaab:
Ok, ready again ☝️

PR #227, opened 2020-01-10
fix(tpi): adjust partners banner to design specs

rapzo:
[PT#2362592](https://www.pivotaltracker.com/n/projects/2362592)


rapzo:
SCSS is a bit messy but bulma's navbar didn't play so well with the design specs. Lots of hammer, lots of nails... Sorry.

PR #231, opened 2020-01-13
Instead of defining height, define width for the homepage logos

aabdaab:
Set a width instead of height, so the height will be calculated automatically. This is a fix for the logos on the TPI homepage.

### DIFF
--- a/app/assets/stylesheets/tpi/_banner.scss
+++ b/app/assets/stylesheets/tpi/_banner.scss
@@ -48,7 +48,7 @@
 
   a {
     img {
-      height: 50px;
+      width: 150px;
     }
   }
 }


rapzo:
Can we wait for my PR to go in? I do some things in the same place. Are the partners dynamic? I adjusted them according to specs' width and height...

aabdaab:
could you paste a code about that bit here? The logos are a bit time-sensitive right now..

rapzo:
Not a very big hit: https://github.com/Vizzuality/laws_and_pathways/pull/227/files#diff-b15b3c357a8a4f3096af295dc926f79a

aabdaab:
hey, let's stick with the commit I added cause setting the height + width is squeezing some logos

PR #232, opened 2020-01-14
Feature/tpi adjust date format

simaob:
Set the default date printing format.


config/initializers/date_time.rb
```diff
@@ -0,0 +1,2 @@
+# Date
+Date::DATE_FORMATS[:default] = '%d/%m/%Y'
```
tsubik: We already have time_date_formats.rb intializer, so better to add that there. Didn't know that this is the way to change defualt format.

tsubik:
When adding new event the date selected by datepicker is still in old format, when editing existing one, it's in the new format. There must be some config for datepicker format.

tsubik:
I think I have an idea how to quickly change the default datepicker format for all date pickers.

config/initializers/date_time.rb
simaob: I'll move it!

simaob:
I see you did it! Thanks =)

PR #233, opened 2020-01-14
Make litigation sides linkable

aabdaab:
This PR adds some polishing to the litigation show page:
* Makes litigations sides linkable (+ styles) if connected entity is present (we don't have an example of that in the DB but there's an existing decorator that takes care of that)
* Add ellipsis to long tag names (labels)

### DIFF
--- a/app/assets/stylesheets/cclow/_geography-page.scss
+++ b/app/assets/stylesheets/cclow/_geography-page.scss
@@ -197,7 +197,12 @@ $count-description-lines: 2;
 
   .details {
     border: solid 1px $grey;
+    color: $blue-dark;
     padding: 3.12rem;
+
+    a {
+      color: $red;
+    }
   }
 
   .climate-targets-section {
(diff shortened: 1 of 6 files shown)


simaob:
I'm not sure if this is really necessary. This is already listed on the right hand side, no? What is the benefit of showing on this sidebar as well? Maybe @faustoperez could share his thought as well. Tks!

simaob:
And another question is the party type related with the sides? Maybe we should show:

**Side A**: Apple (Corporate)
**Side B**: Steve Jobs (Individual)

With link to entities when that's set.

simaob:
(but on the box on the main content).

aabdaab:
In that case maybe we should clarify after all, what is meant by litigation filters in the spreadsheet (row 38).
The second thing, it is on the design so I'll add it.

app/assets/stylesheets/cclow/_geography-page.scss
```diff
@@ -572,10 +577,24 @@ mark {
       color: $blue-dark;
       font-size: 0.75rem;
       line-height: 0.938rem;
+      flex: 1;
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      display: unset;
```
rapzo: Pro move 👏

app/controllers/cclow/geography/litigation_cases_controller.rb
```diff
@@ -17,6 +17,7 @@ def show
         @sectors = @litigation.laws_sectors.order(:name)
         @keywords = @litigation.keywords.order(:name)
         @responses = @litigation.responses.order(:name)
+        @litigation_sides = @litigation.litigation_sides.map { |ls| LitigationSideDecorator.decorate(ls) }
```
simaob: We can't use this decorator directly, because this is for the backend site and the link to Geographies points to the admin namespace. We might need to create one specifically for the frontend, like we have on the Laws side.

PR #235, opened 2020-01-15
Simplifies litigation sides filtering and adds specs

simaob:
When reviewing the filtering PR I completely forgot to review the litigation_sides filtering, sorry about that! 

I have simplified it in this PR and added specs to cover this.

### DIFF
--- a/app/services/queries/cclow/litigation_query.rb
+++ b/app/services/queries/cclow/litigation_query.rb
@@ -114,7 +114,7 @@ module Queries
 
       def litigation_resources(sql)
         ids = ActiveRecord::Base.connection.execute(sql).map { |row| row['litigation_id'] }
-        Litigation.where(id: ids)
+        scope.where(id: ids)
       end
     end
   end
--- a/app/services/queries/cclow/litigation_query.rb
+++ b/app/services/queries/cclow/litigation_query.rb
@@ -111,11 +111,6 @@ module Queries
 
         scope.recent
       end
-
-      def litigation_resources(sql)
-        ids = ActiveRecord::Base.connection.execute(sql).map { |row| row['litigation_id'] }
-        scope.where(id: ids)
-      end
     end
   end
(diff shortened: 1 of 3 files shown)


app/services/queries/cclow/litigation_query.rb
```diff
@@ -204,6 +117,5 @@ def litigation_resources(sql)
         Litigation.where(id: ids)
```
tsubik: One more thing. Here should be `scope.where`, becuase the default scope in this Query object is `Litigation.published`, so to be on the safe side, we cannot show unpublished ones

simaob: Good point. I'll fix it.

simaob: done in a new commit, updated specs as well.

tsubik: Ha, wait! that method was not used anymore? Then it could be deleted ;]

app/services/queries/cclow/litigation_query.rb
```diff
@@ -78,106 +77,20 @@ def filter_by_to_status
...
-
-        sql = "
-          SELECT side_a.litigation_id
-          FROM litigation_sides side_a, litigation_sides side_c
-          WHERE side_a.side_type = 'a' AND side_a.name IN (#{params[:side_a].map { |a| "'#{a}'" }.join(', ')})
-          AND side_c.side_type = 'c' AND side_c.name IN (#{params[:side_c].map { |c| "'#{c}'" }.join(', ')})
-          AND side_a.litigation_id = side_c.litigation_id;
-        "
+          sql += <<-SQL
+            INNER JOIN litigation_sides AS #{side}
```
aabdaab: that's slick 🤔and I was full on with trying to invent ruby's pattern matching for that case 😅

PR #236, opened 2020-01-15
Feature/cclow public downloads

tsubik:
Adding user public downloads for Legislations, Litigations and Climate Targets

Draft as two things could be improved:

- getting all legislations takes like 3 seconds on production data, in staging env. (at least on my machine, we could test that on staging too)
- some simple controller tests could be beneficial, didn't have time for them today

### DIFF
--- a/app/controllers/cclow/climate_targets_controller.rb
+++ b/app/controllers/cclow/climate_targets_controller.rb
@@ -3,6 +3,7 @@ module CCLOW
     include FilterController
 
     # rubocop:disable Metrics/AbcSize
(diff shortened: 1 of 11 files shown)


tsubik:
In development evn. it's gonna take some time.

simaob:
Given we don't have anything yet. We could merge and test on staging.

simaob:
I mean, this is already better than the alternative =D and we can test it there.

tsubik:
Ok!

tsubik:
I'm gonna add some simple tests, and try to make things faster in the meantime.

PR #237, opened 2020-01-15
Feature/quick changes to backend

simaob:
Closes:

- https://www.pivotaltracker.com/story/show/170731182
- https://www.pivotaltracker.com/story/show/170731229

It also adds a filter by Methodology Version to make it easier to download all the MQ Assessments by that filter.


simaob:
Hey @tsubik can you check this one quickly? Should be good I hope! Thanks!

simaob:
True, I'd take that from their use case, let's see how they handle it.

PR #238, opened 2020-01-16
Feature/optimize cclow user downloads

tsubik:
Something weird with AR (when querying Legislation with time range and passing that into CSV exporter), so resetting Active Relation before passing it to CSV exporter objects.

Also, I'm adding some simple controller specs.

### DIFF
--- a/spec/factories/litigations.rb
+++ b/spec/factories/litigations.rb
@@ -35,10 +35,6 @@ FactoryBot.define do
     at_issue { 'At issue Lorem ipsumumum' }
     visibility_status { Litigation::VISIBILITY.first }
 
-    factory :published_litigation do
-      visibility_status { 'published' }
-    end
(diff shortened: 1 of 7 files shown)


spec/factories/litigations.rb
```diff
@@ -35,10 +35,6 @@
     at_issue { 'At issue Lorem ipsumumum' }
     visibility_status { Litigation::VISIBILITY.first }
 
-    factory :published_litigation do
```
tsubik: There are global `traits` in `factories/visiblity_status.rb` file for that.

simaob: ahhhh lame me

tsubik: I guess that's not very intuitive, only finding example could tell that there are some globals.

PR #239, opened 2020-01-16
Add more information to the events on the timeline

aabdaab:
This PR adds additional information to the events that are showing on the timeline (Geography, Legislation, Litigation).
On point hover, we are showing a tooltip with the title of the related `eventable`. 

Screenshot 2020-01-16 at 17 47 43

PT: https://www.pivotaltracker.com/story/show/170645003

### DIFF
--- a/app/controllers/cclow/geography/legislations_controller.rb
+++ b/app/controllers/cclow/geography/legislations_controller.rb
@@ -23,7 +23,10 @@ module CCLOW
         @sectors = @legislation.laws_sectors.order(:name)
         @keywords = @legislation.keywords.order(:name)
         @responses = @legislation.responses.order(:name)
-        @legislation_events = @legislation.events_with_eventable_title
+        @legislation_events = @legislation.events_with_eventable_title.order(:date)
+        @legislation_events_with_links = @legislation_events.map do |e|
+          ::Api::Presenters::Event.call(e)
+        end
       end
 
       private
--- a/app/controllers/cclow/geography/legislations_controller.rb
+++ b/app/controllers/cclow/geography/legislations_controller.rb
@@ -23,8 +23,8 @@ module CCLOW
         @sectors = @legislation.laws_sectors.order(:name)
         @keywords = @legislation.keywords.order(:name)
         @responses = @legislation.responses.order(:name)
-        @legislation_events = @legislation.events.order(:date)
-        @legislation_events_with_links = @legislation_events.map do |e|
+        legislation_events = @legislation.events.order(:date)
+        @legislation_events_with_links = legislation_events.map do |e|
           ::Api::Presenters::Event.call(e, :legislation)
         end
       end
--- a/app/controllers/cclow/geography/legislations_controller.rb
+++ b/app/controllers/cclow/geography/legislations_controller.rb
@@ -23,6 +23,7 @@ module CCLOW
         @sectors = @legislation.laws_sectors.order(:name)
         @keywords = @legislation.keywords.order(:name)
         @responses = @legislation.responses.order(:name)
+        @legislation_events = @legislation.events_with_eventable_title
       end
 
       private
--- a/app/controllers/cclow/geography/legislations_controller.rb
+++ b/app/controllers/cclow/geography/legislations_controller.rb
@@ -23,9 +23,9 @@ module CCLOW
         @sectors = @legislation.laws_sectors.order(:name)
         @keywords = @legislation.keywords.order(:name)
         @responses = @legislation.responses.order(:name)
-        @legislation_events = @legislation.events_with_eventable_title.order(:date)
+        @legislation_events = @legislation.events.order(:date)
         @legislation_events_with_links = @legislation_events.map do |e|
-          ::Api::Presenters::Event.call(e)
+          ::Api::Presenters::Event.call(e, :legislation)
         end
       end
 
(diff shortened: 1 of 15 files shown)


tsubik:
:eyes:

tsubik:
This is good for geography's related events, but we also have `description` on the event table. I see they are not using that field.

tsubik:
Maybe the title of related eventable (only for geography related events) could be shown with smaller font below the event title?

tsubik:
Actually, as a small link to that law/policy/litigation.

tsubik:
On hover there could be more details (if they are using the fields like description/url)

aabdaab:
> Maybe the title of related eventable (only for geography related events) could be shown with smaller font below the event title?

you mean in the tooltip?

tsubik:
Or maybe even previous events for related litigation/policy. Not sure if that could be a helpful for their users.

tsubik:
No, not in the tooltip.

aabdaab:
I also need to add redirection to the related legislation/litigation/geography on the point click.

tsubik:
Tooltips are hard on mobile.

aabdaab:
well, yeah. for a mobile it would have to be a different behavior. but for the desktop, the are some titles that are really long, not sure how that would look like 🤔

tsubik:
We can always trim the text.

tsubik:
Maybe this needs to go back to the design board. But this PR could be the initial version.

aabdaab:
I would say that description on the timeline (on the hover as you're suggesting) would be a bit too much and could get bloated. I see a timeline as just a general visualization of the events. If we add the click behavior to the points, the user will be redirected to the related eventable and could read more about that specific eventable. I would refrain from adding a lot of things in a small space like timeline

aabdaab:
but again, worth checking with @faustoperez

tsubik:
Well, I meant the description that is not shown anywhere on public site. I don't know if they are using that field, if no, then maybe it should be removed.

aabdaab:
I need to check that

simaob:
I'll check in the afternoon sorry, I have some comments about the content of the tooltip but have been in meetings the whole morning.

simaob:
So we have these fields that are relevant:

On the event:

* event type;
* event title;
* description;

On the records:
* title/name.

I think that on the timeline, without the over we should show:

* the date (already there);
* **event title** OR if no event title show **event type** (the description might be too big here, or they might not even be using it)

On hover:

* If it's an event related to the Record we are looking at (so not an "extended event", which we currently show on geographies only) I think we can show the **event description**, because showing the name of the Record doesn't add much; We can show the event title/type if there's no description;
* If it's a node for an extended event, it makes sense to show that eventable title like we have here. Just wondering if we should keep the event type on the tooltip as well, as it gets hidden by the tooltip. But might be overkill, people can "unhover"

aabdaab:
This is ready again ☝️(Event title is a humanized event type, so I'm using event title only)

simaob:
checking.

simaob:
Pretty cool!=) One question, is it easy to have the timeline scrolled to the right on landing?

aabdaab:
to show the latest events? yeah, let me check.

simaob:
Exactly!

aabdaab:
Alright, added the hook for that ☝️

app/controllers/cclow/geography/legislations_controller.rb
```diff
@@ -23,7 +23,10 @@ def show
         @sectors = @legislation.laws_sectors.order(:name)
         @keywords = @legislation.keywords.order(:name)
         @responses = @legislation.responses.order(:name)
-        @legislation_events = @legislation.events_with_eventable_title
+        @legislation_events = @legislation.events_with_eventable_title.order(:date)
```
simaob: @legislation_events doesn't need to be a instance variable, if it's not used in the views.

app/services/api/presenters/event.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      private
+
+      def link(eventable_type, eventable_id)
+        if eventable_type.eql?('Legislation')
+          legislation_route(Legislation.find(eventable_id)&.geography_id, Legislation.find(eventable_id))
+        elsif eventable_type.eql?('Litigation')
+          url.cclow_geography_litigation_case_path(
+            Litigation.find(eventable_id)&.geography_id,
```
simaob: Do we need to pass the eventable_type and eventable_id separately? otherwise we could just have the `eventable` object and from there get to `geography_id`.
Example: if you go to the rails console and get the first event: `e = Event.first` and then access `e.eventable` that will be the correct object: Litigation/Legislation/Target/Geography. So we can make use of those associations, instead of using the find here.

app/services/api/presenters/event.rb
```diff
@@ -0,0 +1,56 @@
...
+          date: event.date,
+          description: event.description,
+          eventable_title: event.eventable_title,
+          link: link(event.eventable_type, event.eventable_id)
+        }
+      end
+
+      private
+
+      def link(eventable_type, eventable_id)
```
simaob: following from the comment bellow we can pass to this method `event.eventable` and then just use that object. Then instead of `eql?` you can use `eventable.is_a?(Legislation)` etc.

app/services/api/presenters/event.rb
aabdaab: true, we don't need to pass anything to these internal methods. I pushed the commit with a simplified code

simaob:
Awesome! =)

PR #241, opened 2020-01-17
feature/tpi-cclow-sort-images

batraisk:


app/views/admin/pages/_image_fields.html.erb
```diff
@@ -4,9 +4,11 @@
   <% if is_new_record %>
     <%= form.input :logo, as: :file %>
     <%= form.input :link %>
+    <%= form.input :name %>
```
simaob: could we put the link and name inputs outside of the if? they seem the same, probably just the image that needs to be inside the if clause.

PR #243, opened 2020-01-20
Feature/update slugs

tsubik:
Change slugs whenever title changes for models with friendly_id module.

I added history for TPISectors and Companies as we are using slugs in the url there. Should be also done for Pages, but that's more work. Maybe instead of Dynamic routing we should have one route `get 'tpi/:page'? I could try that in this PR.

### DIFF
--- a/app/controllers/tpi/companies_controller.rb
+++ b/app/controllers/tpi/companies_controller.rb
@@ -3,6 +3,7 @@ module TPI
     include UserDownload
 
     before_action :fetch_company
+    before_action :redirect_if_numeric_or_historic_slug, only: [:show]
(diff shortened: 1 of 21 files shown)


simaob:
I think that's worth the try!

tsubik:
Ok, After consideration, I think it's better to do smaller steps and do the Page refactor in another PR. Not sure when I can do that as I'm trying to figure out authorization and maybe there are higher priority stories.

tsubik:
So, I'm marking this ready to review.

tsubik:
Also worth to mention that if they already changed some titles, and slugs for them should be updated than we either have to update them manually or they have to change title to previous value and back to normal to get updated slug.

PR #245, opened 2020-01-21
Feature/admin youtube videos

batraisk:


tsubik:
Hmm, I thought this is about allowing to embed youtube videos directly through Trix editor.

tsubik:
Hmm, Trix editor is simple and there are ways to insertHtml into editor, just might not be easy to plug this in our interface. I'm worried that introducing other one like CKEditor could introduce new issues. Although, the client seems to want a way more flexibility in terms of static pages content management, so I'm not sure. Maybe it is worth a try? @simaob

simaob:
I think @aabdaab found it a bit tricky to add the embeds to Trix. But that would be the preferred choice yes. Does trix support other HTML and/or markdown? Those would be nice too.
Adding a different editor at this time might be too much, and bring too many risks.
As an alternative to adding the embeds directly on Trix I suggested adding a different field, for ease of implementation.

tsubik:
It does not support markdown.

app/views/admin/pages/_content_fields.html.erb
```diff
@@ -3,7 +3,11 @@
 <%= content_tag :div, class: 'nested-fields', data: { controller: 'content', new_record: is_new_record } do %>
   <%= form.input :title %>
   <%= form.input :content_type %>
-  <%= form.input :text, as: :trix %>
+  <li class="input">
```
tsubik: I think I can propose some improvement here.

To be able to do something like this

```
form.input :text, as: trix, embed_youtube: true
```

With deafult option to `false`, to not enable this to all trix editor instances across the app.

You can extend TrixInput, it is in `app/inputs/trix_input.rb` and add a controller there if option is present. I think those extra options could be in `options` hash. Take a look at some examples of extending formtastic inputs either in this repo or in active admin, or just on formtastic docs.

app/javascript/admin/controllers/embed_trix_controller.js
```diff
@@ -0,0 +1,52 @@
+import { Controller } from 'stimulus';
+
+const Trix = require('trix');
+
+export default class extends Controller {
```
tsubik: Nice way of extending the editor.

tsubik:
This "enter" problem is when to the fact that Trix attachment is selected (this little close icon is shown). Sometimes the video is like "refreshed" when adding new lines, weird, but not a blocker.

tsubik:
Works great! You could also add this editor to page `description`, and for CCLOWPage and News.

tsubik:
One thing I would think about also is how to make it look better on the page itself. With width: 100% but maintaining aspect ratio.

tsubik:
Something like

```
figure[data-trix-content-type='application/youtube-video.html'] {
  padding-top: 0.8 * 56.25%;
  height: 0;
  position: relative;
  width: 80%;
  margin: 0 auto;

  iframe {
    position: absolute;
	  top: 0;
	  left: 0;
	  width: 100%;
	  height: 100%;
  }
}
```

Placed in `stylesheets/common/youtube.scss` and import that for both TPI and CCLOW. That would make youtube video 80% of the container width with maintaining proper 16:9 aspect ratio.

simaob:
Nice work! =)

PR #246, opened 2020-01-21
Feature/authorization rules ts

tsubik:
This authorization is a pretty complex feature. What if we use validations instead of throwing errors. Here is the PR with suggested changes.

When user will try to publish/unpublish forbidden record he will get a validation error. The same with importing.


tsubik:
Btw. we have an error with data uploading. It was designed to not put through any changes until all validation errors are gone. Meaning, the file with just one row invalid shouldn't update anything. The rollback for changes have been commented out and now it's updating/creating silently in the background the valid rows.

PR #248, opened 2020-01-22
Map changes

aabdaab:
This PR adds changes to the map:
* Change projection to Equal Earth
* Make both the country border and the bubble clickable to reveal the tooltip
* Remove the multiply blending mode on bubbles and z-index order them by value (the bigger would appear first)
* Add option to show European Union's aggregated data
* Add margin-top to the map to avoid being hidden by the map controls

Screenshot 2020-01-22 at 12 52 12


tsubik:
:eyes:

aabdaab:
Regarding the height, there was a requirement about having the whole map module fitting in ~700px of height. Maybe we could sacrifice some paddings in the legend part if the map appears too small 🤔
Distinguishing between mouse click and drag is a good idea, let me tackle that.

tsubik:
Nice :)

Here is one more thing. Sometimes I have weird behaviour, like on the gif below.

weird_behaviour

aabdaab:
this react-tooltip with react-simple-maps produces some weird behaviors. Clicking the geographies to show the tooltip and **at the same time** highlighting its boundaries works inconsistently for some reason. Curiously enough, the issue doesn't reproduce when we don't highlight the strokes (the tooltip always shows in this case when we click on the shape). So, I'm still working on it, checking if the distinction between dragging and clicking has introduced some regressions

aabdaab:
After many turbulences with the different ReactTooltip in the map cases, this is ready again ☝️ Implementation of the new mobile design version will come in a different PR.

tsubik:
Nice! I don't see previous issues anymore. 

Hopefully some small things below. There is a small gray bar at the top of the map, and on country pages, the map is narrower which places checkbox almost at the dropdown. 

image

aabdaab:
thanks for spotting this, it's probably because of the margin-top I added that was meant mainly for the homepage map, let me check this

tsubik:
That margin bar is also on the home page.

aabdaab:
I pushed the commit that removes the grey navbar from the map. About the button overlapping a dropdown, do you have developers tools opened? Or maybe browser zoom is different from 100%? I can't reproduce that:
image

tsubik:
I have 100% set, on firefox I have even checkbox behind the dropdown, hmm. The Width of map is 952px.

tsubik:
maybe that checkbox could go into map__selectors, with flex there and `margin-left: auto` it could be on the right if selectors takes the whole map width. I would set max width for dropdowns.

aabdaab:
ok! could you check again if you still have this issue ☝️

PR #249, opened 2020-01-24
Add Litigation Cases exception for USA

simaob:
Litigation cases from the USA are not managed in this database, so I've added a message to point the users to the right place.

I also removed the filter on links on the profile sidebar when there's
no data for a type of data, as I think this is more consistent and there
might be different reasons to not have data.
Maybe we could add a field to the countries to "Why no Litigation DAta"?
:D but I won't do it here now.

### DIFF
--- a/app/views/cclow/geographies/_overview.html.erb
+++ b/app/views/cclow/geographies/_overview.html.erb
@@ -17,11 +17,7 @@
           <div><%= item[:caption] %></div>
           <div class="indicator-value">
             <div class="value-container">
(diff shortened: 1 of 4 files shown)


simaob:
That's true, but in the future they will be adding some selected cases for the US, so I think it's OK. And people might still be curious "hmmm are you sure there are no cases for the US?"

tsubik:
Haha, yeah, exactly, folks will be like John Cena

PR #250, opened 2020-01-24
Adds validation for geography_id in litigations

simaob:
Small PR to add validation for litigations' geography ID.
I'm not making bigger changes yet, because they still have some records without geography, they'll have to fix that first.

### DIFF
--- a/app/models/litigation.rb
+++ b/app/models/litigation.rb
@@ -119,7 +119,7 @@ class Litigation < ApplicationRecord
     accepts_nested_attributes_for :events
   end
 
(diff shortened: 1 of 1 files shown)


simaob:
Yes, that's true. Maybe that's simpler?

tsubik:
I think removing `optional` it tells right away that it cannot be null.

tsubik:
Otherwise I would think that geography could be null, won't look at validations.

simaob:
changed as discussed.

PR #255, opened 2020-01-28
add cait emissions data

aabdaab:


tsubik:
We also need to enable adding dataset in the code. I think we might change the code to get all datasets we have in the directory. With the use of extra metadata property in yaml file like `show` or `active` we can always disable any of them. `value_column` could be also moved to yaml file. `parse_value` function, hmm, that one could stay or also be in yaml file as text that will be executed  dynamically.

simaob:
This is till WIP, right @edwardpmorris? until we get the data calculated in %?

aabdaab:
this is ready ☝️ As Tomek suggested, we now get all the datasets from `map_data` folder and parse them if they have set `active: true` in the yaml metadata files.

tsubik:
Great :+1:

PR #257, opened 2020-01-28
sort laws and policies by last event date

batraisk:


tsubik:
Hmm, not sure if that is a good idea. I think it could be more complicated. When `q` filter is applied it should be sort by relevance (which I think does not work :/). Otherwise when no `q` filter we can sort by event date. But I guess the user should be notified about sort order.

tsubik:
At some point, something with importing dates must have happened, I have many laws from the future :D. Gonna take a look at develop.

simaob:
Good question, maybe that's the best approach: default sorting by date, but if filters applied (other than date), sort by relevance of search.

PR #258, opened 2020-01-28
use different cut off date then default one when parsing two digit year

tsubik:
I hope that format with the two-digit year is not used by the client anymore :) and it was only for first import (maybe we should actually forbid it at some point)

Although, to make our development data sane and fix it I introduced a different cut off date for a two-digit year format than the default one.

### DIFF
--- a/app/services/csv_import/date_utils.rb
+++ b/app/services/csv_import/date_utils.rb
@@ -1,14 +1,23 @@
 module CSVImport
   class DateUtils
     class << self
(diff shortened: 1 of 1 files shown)


simaob:
Do we need to change anything on the data in prod? I think I fixed that manually at the time, right?

tsubik:
Not sure if they fixed everything. Worth to list all events date and order in console to see.

tsubik:
Some dates were different, so surely they looked at that.

PR #260, opened 2020-01-29
Add button to edit data for logged in users

aabdaab:
This PR adds buttons redirecting to corresponding admin panel sections for logged in user.
Check [PT](https://www.pivotaltracker.com/story/show/170851810) to see in which places the button should appear.

manageDataBTN


tsubik:
Could be useful for TPI side too, pages, news, companies. I'd suggest to do that similarly as in Wordpress. Having fixed top bar and the rest of page should have margin-top. What do you think?

tsubik:
Would be easier as we will only update only one rails template to have this bar. And the link could be provided using variable from controller.

tsubik:
And the user will always know where to expect the link.

aabdaab:
coming back to this :)

aabdaab:
CCLOW fixed navbar:
image

TPI fixed navbar:
image

Not sure about the styles applied so it'd be good to get your opinion about it ☝️ 
The fixed navbar is always visible for logged in users. The title (`Dashboard` on the example) changes based on what page we are in.

app/assets/stylesheets/cclow/layout.scss
```diff
@@ -4,6 +4,10 @@
 
 $search-field-height: 3.75rem;
 
+.header__margin {
+  margin-top: var(--fixed-navbar-height);
```
tsubik: Nice! I have to use those more often

app/controllers/cclow/climate_targets_controller.rb
```diff
@@ -10,6 +10,8 @@ def index
       add_breadcrumb('Search results', request.path) if params[:q].present? || params[:recent].present?
 
       @climate_targets = Queries::CCLOW::TargetQuery.new(filter_params).call
+      @admin_panel_section_title = 'Climate targets'
```
tsubik: I would probably encapsulate this in some common method, similar way as `add_breadcrumb` is done.

app/assets/stylesheets/cclow/layout.scss
aabdaab: yeah! It's really powerful

app/controllers/cclow/climate_targets_controller.rb
aabdaab: good idea

app/assets/stylesheets/cclow/layout.scss
aabdaab: very straightforward article about css variables and how to take advantage of them: https://dev.to/nirazanbasnet/understanding-css-variables-4cj0

tsubik: Thanks!

app/admin/override.rb
```diff
@@ -0,0 +1,5 @@
+ActiveAdmin::Devise::SessionsController.class_eval do
```
tsubik: could be better to add this to active_admin config file in `initializers`

aabdaab:
done ☝️

PR #262, opened 2020-01-29
Avoid showing the future

simaob:
Don't show future publications and articles on the public interface.
Also adds includes clauses to remove N+1 queries.


tsubik:
Should we do the same with Assessments?

simaob:
That's not a bad idea, that way they can add them things when they want.

simaob:
I'll add another task to cover that one.

PR #266, opened 2020-01-31
Feature/tpi sector cluster

tsubik:
- Adding TPI Sector Cluster
- Removing stacked bar borders
- Adding labels with value to individual bars
- Sector Cluster under labels, clusters are sorted by name (not sure if that's ok), sectors withing clusters are sorted by name (SVG madness)
- AxisX labels (sector names) are links to sectors now
- AxisY label `%` added

To import sectors clusters individually run
```
bin/rake reimport:tpi_sector_clusters
```

image


aabdaab:
checking ☝️

aabdaab:
Hey! There's some overhead with the tooltip, I'm taking a closer look.

aabdaab:
Good job with the SVG madness! I've added a commit that changes the plot labels styles to look like on the designs:
image

The performance overhead with tooltips is actually caused by multiple tooltips created by the Bubble Chart. For each bubble, we create a separate tooltip and that affects the whole page. We need to change that to only creating a single tooltip and bind it with an active bubble. I will add that task to PT.

simaob:
noice!

PR #270, opened 2020-02-04
stream publication files until we resolve that differently

tsubik:


tsubik:
That is more a workaround than the real solution though.

tsubik:
Gonna test alternative method of providing direct url to google cloud

simaob:
Cheers, I'll wait for that!

tsubik:
Hmm, I see some problems with just providing direct link to urls. When I attach a new file then the old one is removed from the bucket and the file with new id is uploaded (the old one is still available, weird, not sure for how long). But if someone shares the direct link and then the file will be switched, that's not gonna work as expected. Beside, I don't see filename to be saved as a metadata in google bucket so when downloading the file will have generated unique name, even without extension. That's not good too.

tsubik:
I think we might go with streaming for now.

tsubik:
We might need another controller for uploaded documents.

PR #272, opened 2020-02-05
Avoid crashing tracking activities index page

simaob:
There was a bug getting to the public activities page when you did the following workflow:

- Edit NewsArticle;
- Delete NewsArticle;

We are not treating it as a discardable entity, so when it's gone it's gone, but their activities are still preserved. Maybe we should just cascade delete those activities? But in this PR I'm handling those errors.

Similar for when other entities that are discardable are gone (which might not happen on Prod, but happened locally to me).

If removing the activities is the best way forward, I can add dependent destroy.

Cheers!
Simão

### DIFF
--- a/app/decorators/public_activity/activity_decorator.rb
+++ b/app/decorators/public_activity/activity_decorator.rb
@@ -38,7 +38,9 @@ class PublicActivity::ActivityDecorator < Draper::Decorator
   end
 
   def find_discarded_trackable
-    trackable_type.constantize.all_discarded.find(trackable_id)
+    return nil unless trackable_type.constantize.respond_to?(:all_discarded)
+
+    trackable_type.constantize.all_discarded.where(id: trackable_id).first
   end
 
   def trackable_display_name(trackable_object)


tsubik:
I don't think we should remove activites.

tsubik:
That's our history. Gonna test that.

tsubik:
Hmm, we don't track creates?

simaob:
Good point, I think we might not be tracking those :/ and we should!

Also I'm wondering if we should add "created_by" and "updated_by" to the TPI tables, as that's not set yet.

simaob:
I seem to have broken the specs. Checking!

tsubik:
Deletes are not tracked either, but we won't have any info about the record if it was deleted :/

tsubik:
Ahh, I see it works only with Discardable models.

simaob:
Maybe we should track that something was deleted even if not discardable. If it's not too much work 😓

tsubik:
Hmm, could be some work. To track extra parameter like `title`, which depends on trackable model and render it then on activity list.

tsubik:
https://github.com/chaps-io/public_activity/blob/4f4023d66c92e6950ac517971c399114ebc94dc5/lib/public_activity/roles/tracked.rb#L75

simaob:
I can check it out

tsubik:
but when using `tracked` we need to make sure valid activity is made when discarding.

tsubik:
And change PublicActivity::Common to PublicActivity::Model in the concern.

PR #274, opened 2020-02-13
feature/cclow-sort-litigation

batraisk:


tsubik:
I think the sort code should be moved into the query service. And we should sort by relevance if `q` param was provided. Otherwise we have things like on the screenshot

results-bad

tsubik:
The first item on the list is not even properly visible.

The quick way to sort by relevance would be to add order by id scope to the model 
```
scope :with_id_order, ->(ids) {
    order = sanitize_sql_array(['array_position(ARRAY[?]::int[], id::int)', ids])
    order(order)
  }
```

and then memoize full text search id order and add `merge(order)` at the end of scope chains in query service.

```
      def order
        return scope.with_id_order(@full_text_result_ids) if defined?(@full_text_result_ids)

        # return default order here, so by the last event date
      end
```

tsubik:
@batraisk could you please apply relevance order for Targets, Litigation, Legislations (I think as described above should work) and also default order by last event date.

batraisk:
@tsubik , sure, no problem)

batraisk:
@tsubik , @simaob , I updated the PR. Look please)

tsubik:
Nice! @batraisk can you also add couple tests for testing the order. I'm not sure, those could be either query service tests or controller specs.

batraisk:
@tsubik Done :)

PR #275, opened 2020-02-13
Feature/better cp assessment charts

tsubik:
Well almost done.
new-chart

I'm still missing that fancy labels for benchmark's areas and there is no this dropdown "Showing: Top emitters" and logic related to that. On loading I'm just taking the first 10 for now.

Initially I didn't have this "Apply changes" button, it dynamically changed the selected items. But there was a problem with "Add companies" button and this checkbox list changing position after any change ;].


tsubik:
It's still draft. But could be tested already.

tsubik:
Maybe we can merge this if it's fine and promise to deliver the rest of features later as I won't have much time tomorrow or today to finish the rest.

simaob:
When on a company page this is not behaving as well, maybe because it's using the same component but not adjusted for that? I remove all the items from the legend and then I can't add them back.

Otherwise we could indeed merge. I need to review a little more closely later.

tsubik:
I just hid the remove icon there.

tsubik:
But, found bug when changing assessment. Chart is not rerendering :/

tsubik:
Will take a look at this.

tsubik:
Fixed! ;]

simaob:
Thanks =)

simaob:
Yes, now working. Only "final" remark on this version would be the extra white space under the chart on the Sector page. Which is not in the prod version, hopefully a small tweak and then it should be good as a first major improvement.

tsubik:
I had to use react-highcharts directly, as Chartkick react wrapper wasn't working fine. I think we need to remove chartkick altogether and use react-highcharts everywhere.

tsubik:
Ahh. Right. Good point, I missed that, I'm blind :man_facepalming:

tsubik:
Done.

PR #276, opened 2020-02-18
Improving laws search

tsubik:
I was revising full text search options we are using for all models.

1. Geography
 - removed legislative process from full text search
 - fuzzy matching using trigram algorithm to better match typos

2. Litigations
 - adding side names for full text searching
 - use stemming with english dictionary help
 - ignore accents while searching

3. Legislations
 - use stemming with english dictionary help
 - ignore accents while searching

4. Target
 - removing target_type from search
 - use stemming with english dictionary help
 - ignore accents while searching

Here is also code refactor and UX improvements for search dropdown. I introduced states as `typing` and `searching`, of any of these is happening then search content is in state of `loading` (disabled with opacity). That is eliminating search results "bad states" and empty content while typing.

I also increased search debounce timeout to be 500ms. I think that is reasonable value especially as we have `loading` state now.

new-search

This should be tested on staging first, as I think optimization is not needed yet but we better try if it is performant enough.

### DIFF
--- a/app/assets/stylesheets/cclow/_laws-dropdown.scss
+++ b/app/assets/stylesheets/cclow/_laws-dropdown.scss
@@ -38,7 +38,7 @@ $icon-margin-right: 24px;
   @include desktop {
     padding: 20px 30px;
   }
-  
+
   &[placeholder] {
     text-overflow: ellipsis;
     width: calc(100% - #{$icon-size + $icon-margin-right});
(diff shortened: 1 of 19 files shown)


tsubik:
The most important thing is actually order by relevance. Which is not included here.

simaob:
checking

simaob:
we can add the order by relevant as a separate task, but this is a nice improvement already!

PR #278, opened 2020-02-21
feature/tpi-number-decimal-places

batraisk:


batraisk:
@simaob, I managed to find only two places where numbers with a decimal part are used. If such places still exist, please tell me)

simaob:
Thanks, Andrey!

PR #279, opened 2020-02-21
Feature/cclow sector filter

batraisk:


app/services/queries/cclow/litigation_query.rb
```diff
@@ -43,6 +44,12 @@ def filter_by_region
         scope.includes(:geography).where(geographies: {region: params[:region]})
       end
 
+      def filter_by_sector
+        return scope unless params[:law_sector].present?
+
+        scope.includes(:laws_sectors).where(laws_sectors: {id: params[:law_sector]})
```
simaob: we can just use `joins` here, if we don't need to include the laws_sectors objects in Ruby.

app/services/queries/cclow/legislation_query.rb
```diff
@@ -45,6 +46,12 @@ def filter_by_region
         scope.includes(:geography).where(geographies: {region: params[:region]})
       end
 
+      def filter_by_sector
+        return scope unless params[:law_sector].present?
+
+        scope.includes(:laws_sectors).where(laws_sectors: {id: params[:law_sector]})
```
simaob: same as below

PR #283, opened 2020-02-26
update ruby version and fix travis

tsubik:


tsubik:
Is the ruby update easy when deploying to the server?

tsubik:
at least on cloud66  which is also using capistrano, updating ruby is not that pleasant (can't rollback later ;))

simaob:
We are using rvm, so I might need to go to the server install that new version and only afterwards deploy it. Not sure if it requires a change to Apache config files too, as in my own server I had to change the ruby path

tsubik:
Hopefully, it will be fine.

simaob:
I added the rvm config string to the capistrano config.

PR #285, opened 2020-02-26
Fix queries

tsubik:
This PR is fixing queries to properly order by relevance or last event date. Fixing the duplicated records after filtering (as there was no global distinct and `joins` are being used).

Also:
- Fixing duplicates in instruments filter

Gonna fix line clamp in separate PR.


simaob:
checking

PR #288, opened 2020-02-26
Updates format to match the right format

simaob:
We were not using the right output format for the Climate Targets API endpoint used by Climate Watch, this updates it. We are aiming to match this: http://www.lse.ac.uk/GranthamInstitute/wp-json/wri/v1/targets/BRA

### DIFF
--- a/app/controllers/cclow/api/targets_controller.rb
+++ b/app/controllers/cclow/api/targets_controller.rb
@@ -13,24 +13,22 @@ module CCLOW
       private
 
       def targets_as_json(geography)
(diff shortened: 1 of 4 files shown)


simaob:
weird, let me check what's wrong. maybe I'm assuming there will be data.

simaob:
moved code out, and forgot to take the method with me 😓

simaob:
moved the method to the legislation model, and changed a spec so that it would have caught this

simaob:
I think I broke Github here. The changes are merged, but the app is confused. I'll close the PR.

PR #289, opened 2020-02-27
Feature/country profile

simaob:
This PR adds three new fields for Geographies metadata, the data file to be used and an import script to import the data. It also enables editing on the backend.

You'll need to migrate the database: `bundle exec rails db:migrate`
And then you can import the data with: `bundle exec rails reimport:geographies_metadata`

### DIFF
--- a/app/admin/geographies.rb
+++ b/app/admin/geographies.rb
@@ -11,7 +11,8 @@ ActiveAdmin.register Geography do
 
   permit_params :name, :iso, :region, :federal, :federal_details,
                 :legislative_process, :geography_type,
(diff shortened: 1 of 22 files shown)


tsubik:
Hmm, we can start with that, although not sure if that data should be editable at the backend. I would just add that csv file to the repo and load from there.

tsubik:
Or proxy that from external API.

simaob:
Some of the data is available on Climate Watch, but not on specific endpoints, and the other data is in downloads from the WB... :/ Not ideal, but I'd start with this for the sake of time.

PR #291, opened 2020-02-27
Feature/change publication date to datetime

simaob:
As requested:

- changed publication date from date to date time, so that they can use the time to tweak the sorting of publications;
- removed article type from NewsArticles;


simaob:
rebased with develop

simaob:
ah, probably, let me try that. When I searched for it, I didn't find this link, just an older one to use the formtastic inputs. Let me try this one.

simaob:
Yes, that works! =D thanks for spotting it

simaob:
I pushed the change, but I can't see it here :/

PR #292, opened 2020-02-27
[CCLOW] Style static pages according to UI kit

aabdaab:
[CCLOW] This PR adds more accurate styling to static pages according to UI kit:
- bullet lists;
- numbered lists;
- headings;
- bold;
- links

### DIFF
--- a/app/assets/stylesheets/cclow/_pages.scss
+++ b/app/assets/stylesheets/cclow/_pages.scss
@@ -2,6 +2,9 @@
 @import "colors";
 @import "bulma/sass/utilities/all";
 
+$text-indent: 25px;
+$text-indent-numbers: 18px;
+
 .pages__title {
   margin-top: 40px;
   margin-bottom: 40px;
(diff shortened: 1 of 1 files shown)


simaob:
Not sure if this is like this on the UI kit, but I wonder if the bullets and numbered lists can be a little better aligned with the titles, they feel a bit out of position:

Screenshot 2020-02-27 at 13 39 07

aabdaab:
On the designs they are outside of the grid: https://projects.invisionapp.com/d/main#/console/18784734/392404645/inspect
it does look a bit confusing though..

simaob:
oh yes, you're right. Maybe that's just how it is! =) @faustoperez can you confirm? =) (see screenshot above).

tsubik:
I agree that looks quite off

faustoperez:
Yes, it is off on purpose on the designs (it's an actual recommendation on a Typography book I read), but please use the standard user agent margins, I don't have a hard preference and this seems to put people off :). Thanks!

aabdaab:
Thanks for the input! I aligned the list items with the grid ☝️

PR #297, opened 2020-03-01
Fix cp chart

tsubik:
Fixing CP chart problems.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -46,6 +46,8 @@ rules:
   react/jsx-indent: 1
   react/jsx-wrap-multilines: 0
   react/prefer-stateless-function: 0
+  react-hooks/rules-of-hooks: error
+  react-hooks/exhaustive-deps: warn
 parser: babel-eslint
 parserOptions:
   ecmaVersion: 7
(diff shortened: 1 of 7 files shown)


simaob:
I'm seeing something a little weird with the pledges:

Screenshot 2020-03-02 at 11 49 37
Screenshot 2020-03-02 at 11 49 46


When changing the assessment date, the pledges numbers are very different. Is this a data issue or could it be the chart misbehaving?

tsubik:
The unit is different, the benchamark is different.

tsubik:
It is always taking the right benchmark comparing benchmark release date with assessment publication date.

tsubik:
Hmm, the unit is the same.

simaob:
but am I seeing it wrong I see: gCO2/passenger-per-km

simaob:
exactly! I had to triple check.

tsubik:
Maybe they didn't update the unit?

simaob:
I'm looking at local data, so might be just wrong data. I'll check on staging!

tsubik:
On prod, there is also the same unit for both. Other unit, but there should be different units.

simaob:
Maybe they haven't used the new feature of adjusting units ...

PR #300, opened 2020-03-04
implement mobile versions of the Search page

batraisk:


simaob:
I think I see a problem with closing a tag, once it's selected, the interface doesn't seem to update on mobile, but it does in the desktop version.

simaob:
e.g.: selected Chile as a country, then click the little `x` and the tag stays there. Testing in responsive mode in Google Chrome.

batraisk:
I checked and I did not have such a problem. I'll try to figure it out :)

batraisk:
@simaob , I found the cause of the problem and fixed it.

PR #301, opened 2020-03-09
More chart changes

tsubik:
- Removing chartkick and moving all charts to JS
- Company level chart changes to look more like on designs
- All sectors chart: changing tooltip, white data label for "No Disclosure"

mq-level
changes-to-cp-all-sectors


tsubik:
I didn't even notice they weren't sorted properly.

PR #303, opened 2020-03-16
System tests

tsubik:
Introducing a couple of simple system tests for public pages. I was thinking if to use Cypress or just capybara for it and for now I decided to go with Capybara.

The tests are run against the database restored from development database dump. To recreate the development database dump just invoke command `bin/rake test:db_dump`. That will clear the test database, load the seeds and then dump the whole database into the file and clear it again. I'm wondering if not just use factory bot for this, but I was lazy and tested against development data.

The problem here is that development data is not great, and it needs to be improved as there are some missing relations between models and not every part of the system have any example data (maybe by seed_dump of part of production data). 

I set the version of `rspec-rail` to `4.0.rc` version as this one seems to fully support Rails6. On the previous version failed test screenshots were blank.

I also fixed some links here and there to use slugs and not `id`. LawsSector needs `slug` too as climate targets per LawsSector url should be improved, but that could be done in separate PR.


spec/system/public/cclow/climate_targets_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+describe 'Climate Targets search', type: 'system' do
+  before do
+    visit 'cclow/climate_targets'
+  end
+
+  it 'loads the page' do
+    expect(page).to have_text('All Climate Targets')
+    expect(page).to have_text('Showing 1993 results')
```
simaob: These tests would need to be updated if we change the test dump, right? Not sure if I'm understanding correctly.

tsubik: Yes, that's why it's not that great. Although, seeds were not changed for a quite a while

PR #306, opened 2020-03-23
Updating csv files

tsubik:
Updating Litigation and Legislation public format to the desired new structure.

- testing CSV files using snapshots (similar to Jest snapshot, using rspec-snapshot library). Easier to check if all data is exported in the correct way. The only drawback is to eliminate all dynamic values, sequences etc. otherwise every time the test will run the snapshot won't match
- added order by geography name for litigation. legislations and targets public CSV files
- getting all legislations csv might be too slow now :/, maybe the next step will be to introduce database view to get the data.

Should be tested on staging first.


tsubik:
See the tests are still failing. Gonna fix that first.

tsubik:
I totally don't get it why returned tags which are just `has_many` relation are in non-deterministic order. There should be in the same order as they were added, but sometimes they are, sometimes they don't :/

I'm trying to understand that.

simaob:
Are they just random? :/ maybe use a non order specific matcher?

tsubik:
snapshot_matcher is matching the whole file content. That was the point of using it. Running the test for the first time, looking at the file using installed spreadsheet app, checking if everything is ok.

tsubik:
Don't want to replace it with multiple matchers :/

tsubik:
I could order by in the relation, but I thought it should just work.

tsubik:
I think in the next PR I'm gonna remove those rubocop metrics like AbcSize and MethodLength, we are only disabling and enabling that LOL. That's annoying.

simaob:
good idea, it gets a bit annoying. It's quite hard to stick to them :s

tsubik:
Looks like tests are passing, but I still don't understand that non-deterministic issue ;] maybe because of joins postgres is sometimes returning in different order.

tsubik:
Yeah, in most cases we might use `let_it_be` or just before_all with database cleaning afterwards. But, that must be done gradually.

PR #308, opened 2020-03-27
Fixing domain constraints

tsubik:
Still as a draft as I need to test it more.

Also, would be great to test that on staging. Maybe we can set 2 domains/subdomains for staging cclow and tpi.

Locally everything works just fine, to access tpi website `tpi.localhost:3000` and to access cclow `cclow.localhost:3000`. Of course, after adding entries to `/etc/hosts` this is described in readme.


simaob:
I'm going to try to get the staging domains.

config/routes.rb
```diff
@@ -1,80 +1,87 @@
...
+      root to: 'home#index'
 
-    get '/about', to: 'home#about'
-    get '/newsletter', to: 'home#newsletter'
-    get '/register', to: 'home#register'
-    get '/disclaimer', to: 'home#disclaimer'
-    get '/sandbox', to: 'home#sandbox' if Rails.env.development?
+      # maybe temporary redirects to remove /tpi part
+      get '/tpi', to: redirect('/')
+      get '/tpi/*rest', to: redirect('%{rest}')
```
tsubik, line 8: those redirect fallbacks should just work @simaob

tsubik:
Ohh crap, we use develop not master only. uhh

PR #309, opened 2020-03-27
some targets are without sectors, add includes for quicker download

tsubik:
Fixing the latest error with exporting targets.


tsubik:
rebased

PR #311, opened 2020-03-27
Linking to the search results page from country page

simaob:
OK, I'm a little stuck here, can you help me Tomasz? =)

We have decided to link the Country page sidebar directly to the search results page so that users can filter the data further, specially for cases, like Australia, that have a lot of Litigation Cases, but as we shouldn't do it just for some, we are doing for all of them.

This first bit was easy, the components already filter the data correctly, but the interface is not in sync with the params, so it's not good. What is needed now is that the search filters are started with an already set state. But I don't know how best to do it, without hacking something nasty! So would really appreciate your help! Tks!

### DIFF
--- a/app/assets/stylesheets/cclow/_filters.scss
+++ b/app/assets/stylesheets/cclow/_filters.scss
@@ -106,6 +106,10 @@ $time-slider-right-handle-width: 18px;
         &:not(:first-child) {
           margin-top: 1rem;
         }
+
+        label {
+          cursor: pointer;
+        }
       }
     }
   }
--- a/app/assets/stylesheets/cclow/_geography-page.scss
+++ b/app/assets/stylesheets/cclow/_geography-page.scss
@@ -617,7 +617,9 @@ $count-description-lines: 2;
   }
 
   .filter-column {
-    margin-top: 1.5rem;
+    @include from($desktop) {
+      margin-top: 1.5rem;
+    }
 
     .search-by {
       color: $grey-text;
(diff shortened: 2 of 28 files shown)


tsubik:
Alright, I will take care of this. So, the state of filters should be taken from query string.

tsubik:
I think it's a chance to refactor those 3 similar lists components to reuse most of the code across them. So, little more patience required on this story ;], but it I think it already works fine on LegislationAndPolicies component.

tsubik:
Syncing all filter params from querystring, so it will be possible to share the results link with others later.

tsubik:
TODO checklist:
- [ ] refactor list components to reuse common code
- [ ] adjust filter mobile designs
- [ ] fix geographies filter - change AND to OR

tsubik:
What to do if there are no connected entities?
image

simaob:
oh right... maybe the same message we have currently on the page. since it's replacing that.

tsubik:
I don't think the same message is appropriate as those are search results. Should be more universal like "No litigation cases match your search criteria."

tsubik:
This is done @simaob  :)

simaob:
Given that people are linked from the country page, I'd try to use the referrer to make the message more specific in that situation. And then use a generic for everything else. What do you think?

Because they are in country context, click "Litigations" (for China) and see nothing.

simaob:
gonna check it now, thanks!

tsubik:
Ok. I will try to do that. So, the same message? What if they change any of the filters? Maybe if only geography filter is changed then back to original message, although I don't like to complicate things lol

simaob:
hmmm if request.referrer == geography_path -> that message, if they change any filter, the current message? Or is that too hacky? :D

tsubik:
I need to implement this to see :]

tsubik:
Ok done.

PR #314, opened 2020-04-08
feature/tpi-pages

batraisk:


batraisk:
@simaob , Please, check it. If this design option does not fit, then I will ask for the help of Fausto)

simaob:
I think this looks OK =) I will see how it fits with the production data!

PR #319, opened 2020-04-22
Chore/update dependencies

tsubik:
Seems like activeadmin in version 2.7 supports webpacker. We already are using webpacker for js assets in activeadmin with some hacky way, but maybe it would be great to revisit that workaround in some time.


tsubik:
I didn't update all js dependencies to the latest versions as upgrading eslint stuff or babel is always a nightmare ;]

PR #320, opened 2020-04-22
Date filter updates

tsubik:
Date filters update

- changing time range filter title to `Date of last change` for both Litigations and Laws
- fixing time range filter sync with query string (it does not work on prod, it's not showing "x Selected"
- adding new time range filter for Laws and Policies (Date of law passed, or should I name it `Date law passed`?)

Should I also add filter for litigation opened events? How to name it if I have to add it? 

image


tsubik:
I'm going to add that extra filter for Litigation cases, why not, it should be quick.

tsubik:
Already started lol.

simaob:
Alright =) do it, it makes sense. I will check the PR later today or tomorrow, need to wrap up some other things in the meanwhile!

tsubik:
image

I think this "2 selected", "1 selected" here for TimeRange filter does not make any sense. I will change it to: "From 2013 to 2019", "From 2013", "To 2019"

simaob:
yeah you're right, makes no sense.

PR #322, opened 2020-04-30
Updates the available event types for legislation

simaob:
This PR updates the list of event_types for legislation, based on the updated data provided by the client.

The migration files will not fit with this data, but I won't worry too much about it now, as we could always download the updated dataset from prod if needed in the future.

### DIFF
--- a/app/models/legislation.rb
+++ b/app/models/legislation.rb
@@ -33,31 +33,10 @@ class Legislation < ApplicationRecord
   LEGISLATION_TYPES = %w[executive legislative].freeze
   EVENT_TYPES = %w[
     amended
-    approved
-    deadline_for_regulation
-    decree_passed
-    document_amended
-    document_passed
-    document_approved
-    endorsement
-    entry_into_force
-    executive_decree_issued
-    federal_decree_issued
-    first_phase_approved
-    last_amended
-    last_amendment
-    law_amended
-    law_passed
-    law_published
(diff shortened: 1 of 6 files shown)


app/models/legislation.rb
```diff
@@ -130,7 +109,7 @@ def events_with_eventable_title
   end
 
   def date_passed
-    events.where(event_type: 'law_passed').first&.date
+    events.where(event_type: 'passed/approved').first&.date
```
tsubik, line 112: why not just `passed.first&.date`

tsubik:
Hmm, maybe migration script? That could broke existing data.

app/models/legislation.rb
simaob: I don't know :D I just changed the event type :P but good catch, I'll update it =)

tsubik:
Not sure if user will be able to save Law if existing event types are different than the one in enum.

simaob:
I have an upload file that should update ALL of the legislation's events.

tsubik:
Have to go shopping, be back in 1 hour.

simaob:
Yes it worked well locally =)

simaob:
tks!

simaob:
shoot, after I merged I noticed that something is amiss with the filter for law passed with the updated data. I'll check it in develop.

PR #422, opened 2023-07-03
Show multiple charts with common legend

Bluesmile82:
image

Multiple charts are shown on the Carbon performance emission graphs per individual sectors

The shared legend is shown at the top.

https://vizzuality.atlassian.net/browse/LSE-84?atlOrigin=eyJpIjoiZTY5NTc3OThmNTkxNDkyZTg3MmNiYzAzZTAwZmE1ZjAiLCJwIjoiaiJ9

### DIFF
--- a/app/assets/stylesheets/tpi/charts/cp-performance.scss
+++ b/app/assets/stylesheets/tpi/charts/cp-performance.scss
@@ -212,3 +212,51 @@
     }
   }
 }
+
+.charts-container {
+  display: grid;
+  grid-template-columns: 1fr;
+
+  @include desktop {
+    grid-template-columns: 1fr 1fr;
+  }
+  grid-gap: 20px;
+}
+
+
+.chart-container {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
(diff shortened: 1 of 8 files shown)


martintomas (changes requested):
I will not comment code, because I am not expert at this :). Anyways, I have tested it on my local machine and I think it looks overal good, just a few points:

- not really sure if we need legend per every graph, but I think this is up to Teona to decide
- when you resize screen, button called `See full sector chart` pops up. We need to get rid of it for Banks
Screenshot 2023-07-04 at 9 38 04

- When I open console, I see couple errors related to missing `unit` property. We should probably get rid of this error. Also some sectors don't have unit, but some of them have (for example unit of Steel sector is `Carbon intensity (tonnes of CO2 per tonne of steel)`). If possible, we should show such information somewhere at graph.

Screenshot 2023-07-04 at 9 37 39

Bluesmile82:
Lots of thanks for the review. I added the unit and removed the sectors for this case
image

martintomas (approved):
👍

PR #424, opened 2023-07-05
Feature/cp matrix

Bluesmile82:
Add cp matrix table with tabs and legend

image

### DIFF
--- a/app/javascript/components/tpi/charts/cp-matrix/cp-matrix-table.jsx
+++ b/app/javascript/components/tpi/charts/cp-matrix/cp-matrix-table.jsx
@@ -10,7 +10,7 @@ import chevronIcon from 'images/icon_chevron_dark/chevron_down_black-1.svg';
 const COLORS = {
   'Not assessable using TPI’s methodology': { color: '#CACBCE' },
   'No or unsuitable disclosure': { color: 'white', border: true },
-  'Not aligned': { color: '#ED3D4A' },
+  'Not Aligned': { color: '#ED3D4A' },
   'National Pledges': { color: '#F6C242' },
   'Below 2 Degrees': { color: '#F9DF65' },
   '1.5 Degrees': { color: '#57BE77' }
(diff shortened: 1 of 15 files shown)


app/javascript/components/tpi/charts/cp-matrix/cp-matrix-table.jsx
```diff
@@ -0,0 +1,133 @@
+/* eslint-disable function-paren-newline */
+/* eslint-disable implicit-arrow-linebreak */
+import React from 'react';
+import PropTypes from 'prop-types';
+import Table from 'rc-table';
+import ReactTooltip from 'react-tooltip';
+import cx from 'classnames';
+import chevronIcon from 'images/icon_chevron_dark/chevron_down_black-1.svg';
+
+const COLORS = {
```
martintomas, line 10: List of available names can be seen at https://github.com/Vizzuality/laws_and_pathways/blob/develop/app/models/cp/alignment.rb#L27. Based on this list, `Not aligned` should be `Not Aligned`. I will also add `Not assessable using TPI’s methodology` on BE side to this list.

martintomas (changes requested):
Looks good to me, just two minor points:
 - matrix is not responsive, it is not issue for matrix itself, but tabs and legend should change appropriately. Maybe it is not part of this PR
 - I see kind of artefact at right border of matrix (probably for scrollbar). I would be great to get rid of it, but no big deal
Screenshot 2023-07-06 at 9 37 25

app/javascript/components/tpi/charts/cp-matrix/cp-matrix-table.jsx
martintomas: I have covered this with my commit ;)

martintomas (approved):

Bluesmile82:
Hey! Thanks for the review! I have updated it to be responsive. I can't reproduce the artifact in Chrome, Firefox or Safari though

PR #427, opened 2023-07-19
Feature/scroll hint

Bluesmile82:
For the table add a scroll hint that disappears when the user scrolls and always show the scrollbar


Bluesmile82:
Hi @martintomas I see the tests are failing but I don't see anything related with the changes. Could you check this?

PR #451, opened 2023-10-04
Create countries bubble chart

barbara-chaves:
This PR created the countries bubble chart

[Design](https://www.figma.com/file/ZyYNWRf3SJsJaBY4R9bmen/TPI---1280-%5BInternal%5D?type=design&node-id=1936-13169&mode=dev)

### Testing
Go to '/ascor' 
The bubble chart should be displayed


martintomas (changes requested):
Looks good to me ;). There are just two things which I have discovered:
 - the most serious is that `market_cap_group` is not taken into account --> size of the bubble is always same, it should depend on it
 - console is full of some errors such as `Chart.js:36 WRONG INDEX {pillar: 'Climate Finance', area: 'Renewable Opportunities', result: null, country_id: 15, country_name: 'Thailand', …}`
 - also PR contains code for Companies Accordion, but I don't think it is used anywhere <-- if it is true, better to remove it now

barbara-chaves:
> Looks good to me ;). There are just two things which I have discovered:
> 
> * the most serious is that `market_cap_group` is not taken into account --> size of the bubble is always same, it should depend on it
> * console is full of some errors such as `Chart.js:36 WRONG INDEX {pillar: 'Climate Finance', area: 'Renewable Opportunities', result: null, country_id: 15, country_name: 'Thailand', …}`
> * also PR contains code for Companies Accordion, but I don't think it is used anywhere <-- if it is true, better to remove it now

@martintomas 
I added the market_cap_group scales, but it might need tweaking when it comes to the real data
I have nothing in my console, but it looks like the data you have is different from mine (you have  'result: null' in your data) 
I deleted the Companies Accordion component

PR #455, opened 2023-10-09
Add filters

barbara-chaves:
This PR adds the countries emissions chart

For the first review I prepared a mocked data to make the chart data visible for testing. After conforming that the functionality and styles are correct, I will change it to the real data before merging. 

[Figma design](https://www.figma.com/file/ZyYNWRf3SJsJaBY4R9bmen/TPI---1280-%5BInternal%5D?type=design&node-id=1936-13169&mode=dev)

## Testing instructions 

1. Go to '/ascor'. You should be able to see the line chart in the 'Country emission pathways' section
2. The chart data is mocked, so the filters won’t work, but you can test changing the value
3. The plot style, the lines style (targeted vs  historical data), the tooltip and the legend should respect the design


martintomas (approved):
Looks great, thanks @barbara-chaves . Just three minor things which I don't think block this from merge -- feel free to do it anytime ;)
 - I personally don't like behaviour of that country selector --> how picked countries are at top and everytime when you select/unselect new country, list reorders. I would probably order countries by name and that is it...
 - list of selected countries under emissions graph does not get updated based on picked countries at selector
 - both filters and country selector does not fit on mobile screen (it is not responsive)

PR #461, opened 2023-10-17
Add mobile version of the countries emissions bubble chart

barbara-chaves:
This PR adds the mobile (devices smaller than 992px) version of the ascor countries bubble chart


martintomas (approved):
I have tested this and looks good to me ;). Just two minor comments:
 - maybe it would be nice if you could click on country name and it would redirect you to country detail (same way that allows you to click on bubble at bubble chart)
 - It feels to me that orientation of dropdown arrows is reversed

barbara-chaves:
> I have tested this and looks good to me ;). Just two minor comments:
> 
> * maybe it would be nice if you could click on country name and it would redirect you to country detail (same way that allows you to click on bubble at bubble chart)
> * It feels to me that orientation of dropdown arrows is reversed

You are right! I made the changes you suggested. Thanks

